PR #20, opened 2026-05-07
feat: add rhdh-pr-review skill for testing operator PRs on live clusters

Fortune-Ndlovu:
Automates the PR review workflow: fetches CI-built images from GH Actions comments, ensures a running RHDH cluster (provisioning via rhdh-test-instance if needed), swaps the operator image (CSV patch for OLM, oc set image for non-OLM), and generates a targeted review checklist from the diff.

NOTE: we are starting with the operator we will follow up on the helm-chart, rhdh-local, etc, thanks!


schultzp2020:
I'll take a look!

skills/rhdh-pr-review/SKILL.md
```diff
@@ -0,0 +1,83 @@
+---
+name: rhdh-pr-review
+description: Test PR changes on a live RHDH cluster. Fetches CI-built images from PR comments, checks cluster status (deploying if needed), swaps images into the running deployment, and generates a targeted review checklist from the diff. Currently supports rhdh-operator PRs.
```
schultzp2020: Description lacks trigger phrases. Per the [Agent Skills spec](https://agentskills.io/specification), the description is the only triggering mechanism — agents see it at startup to decide whether to activate the skill. Right now it describes what the skill does but doesn't include "Use when..." phrases, so it likely won't auto-trigger on natural language requests like "help me review this operator PR" or "test this PR on my cluster."

```suggestion
description: Test PR changes on a live RHDH cluster. Fetches CI-built images from PR comments, checks cluster status (deploying if needed), swaps images into the running deployment, and generates a targeted review checklist from the diff. Use when asked to review an rhdh-operator PR, test PR changes on a cluster, swap CI images, or deploy PR images for testing. Also use when user mentions "operator PR", "review PR", or "test this PR on my cluster". Currently supports rhdh-operator PRs.
```

skills/rhdh-pr-review/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+<principle name="use_pr_comment_images">
+Extract image URLs from PR comments posted by CI — never construct image URLs manually.
+The tag format includes PR number + commit SHA, which only CI knows.
+See `references/operator-pr-images.md` for extraction commands.
+</principle>
+
+<principle name="patch_not_redeploy">
+Swap images on a running deployment rather than redeploying from scratch.
+Use `oc set image` to patch the operator deployment in place — this is faster and preserves existing state (Backstage CRs, config, Keycloak).
+</principle>
```
schultzp2020: The `patch_not_redeploy` principle says to use `oc set image`, but the workflow (Phase 4.4a) correctly patches the CSV for OLM-managed installs — because OLM owns the Deployment and will overwrite direct changes. Is it worth updating this principle to cover both paths?

```suggestion
<principle name="patch_not_redeploy">
Swap images on a running deployment rather than redeploying from scratch.
For non-OLM installs, use `oc set image`. For OLM-managed installs, patch the CSV — never patch the Deployment directly as OLM will overwrite it. This is faster and preserves existing state (Backstage CRs, config, Keycloak).
</principle>
```

skills/rhdh-pr-review/SKILL.md
```diff
@@ -0,0 +1,83 @@
...
+
+<skills_index>
+
+| Skill | Purpose | Path |
+|-------|---------|------|
+| rhdh | Orchestrator, environment status, activity tracking | `../../rhdh/SKILL.md` |
+
+</skills_index>
+
+<success_criteria>
```
schultzp2020, line 83: Success criteria is defined here (5 items) and again at the end of `workflows/review-operator-pr.md` (8 items, more detailed). The workflow version is a superset. Would be cleaner to single-source it in one place and reference from the other so they don't drift apart.

skills/rhdh-pr-review/workflows/review-operator-pr.md
```diff
@@ -0,0 +1,503 @@
...
+
+```bash
+skopeo inspect docker://quay.io/rhdh-community/operator:<tag> --raw 2>/dev/null
+```
+
+If validation fails, warn that images may have expired (14-day TTL).
+
+---
+
+## Phase 3: Ensure a Running RHDH Cluster
```
schultzp2020, line 66: Phase 3 (cluster provisioning) runs ~130 lines and is only needed when the user doesn't have a cluster or RHDH instance. Would it make sense to extract it to `references/cluster-provisioning.md` with a conditional pointer: "Read `references/cluster-provisioning.md` if no RHDH operator is running"? That'd bring the workflow under ~370 lines and make the provisioning steps reusable if other workflows need them later.

skills/rhdh-pr-review/workflows/review-operator-pr.md
```diff
@@ -0,0 +1,503 @@
...
+  --head <your-fork-branch>
+```
+
+**Step 2: Trigger deployment**
+
+```bash
+TEST_PR=<pr-number-on-rhdh-test-instance>
+
+gh pr comment $TEST_PR --repo $TEST_INSTANCE_REPO \
+  --body "/test deploy operator 1.9 4h"
```
schultzp2020: `1.9` will go stale as new versions release. Consider adding a comment to determine the current version dynamically.

Maybe: `# Replace 1.9 with the current RHDH minor version — check rhdh-operator Makefile VERSION`

Same applies to lines 221 and 224 where `VERSION=1.9` is hardcoded.

skills/rhdh/SKILL.md
```diff
@@ -145,12 +151,20 @@ What would you like to do?
 
 **To route:** Read `../rhdh-local/SKILL.md` and follow its intake process.
 
+### Review & Testing Routes
+
+| Response | Skill |
+|----------|-------|
+| 8, "review PR", "rhdh-pr-review", "test PR", "operator PR", "swap image" | Route to `@rhdh-pr-review` skill |
```
schultzp2020, line 158: The other sub-skill routes use file paths like `../overlay/SKILL.md`, but this one uses `Route to @rhdh-pr-review skill`. Was the `@` prefix intentional as a new convention, or should this match the existing pattern for consistency?

tests/unit/test_review_pr_skill.py
```diff
@@ -0,0 +1,175 @@
+"""Tests for review-pr skill structure and content validation."""
```
schultzp2020: Are these tests necessary? Skills are meant to be iterated on quickly — structural tests that assert specific XML tags and sections exist can become friction when you want to restructure the skill. Is there a case where these tests would catch something that a quick `/skill-maker` audit wouldn't?

schultzp2020 (commented):
Nice work, Fortune! A few questions and suggestions inline.

skills/rhdh-pr-review/SKILL.md
```diff
@@ -0,0 +1,83 @@
+---
```
schultzp2020, line 1: Before merging, it'd be worth running `/skill-maker` in audit mode against this skill. It checks frontmatter, structure, progressive disclosure, reference architecture, and script patterns against the [Agent Skills spec](https://agentskills.io/specification). It would catch things like the missing trigger phrases in the description, the duplicate success criteria, and flag the workflow length.

> `/skill-maker` → select "audit" → point it at `skills/rhdh-pr-review/SKILL.md`

You'll need to install the `skill-maker` skill first if you haven't already.

Fortune-Ndlovu: thanks @schultzp2020 , yeah, another thing worth noting is, this skill is very much operator opinionated and not ANY pr agnostic or at least the helmchart repo also, it might make sense to do the sub skill approach you mentioned earlier

skills/rhdh-pr-review/SKILL.md
Fortune-Ndlovu: added trigger phrase, thanks!

skills/rhdh-pr-review/SKILL.md
Fortune-Ndlovu: principle should document both OLM (CSV patch) and non-
OLM (oc set image) paths, thanks!

skills/rhdh-pr-review/SKILL.md
Fortune-Ndlovu: single-sourced to the workflow's 8-item checklist, thanks!

skills/rhdh-pr-review/SKILL.md
schultzp2020: I have a few ideas on how we can make this skill more generic. I'll ping you with the PR with some additions after we merge this PR!

skills/rhdh-pr-review/workflows/review-operator-pr.md
Fortune-Ndlovu: sounds great, ive extracted Phase 3 to cluster-provisioning.md

skills/rhdh-pr-review/workflows/review-operator-pr.md
Fortune-Ndlovu: agreed

skills/rhdh/SKILL.md
Fortune-Ndlovu: The tests catch  regressions when someone restructures the skill and forgets a required section. They also served as TDD guardrails during initial development. But yeah you have a point about friction.

tests/unit/test_review_pr_skill.py
Fortune-Ndlovu: good call removed them.  /skill-maker audit covers the same ground without creating friction during  iteration.

skills/rhdh-pr-review/SKILL.md
Fortune-Ndlovu: FYI audit against the skill maker passes e.g.:

  Basics                                                                       
                                                                               
  - name is lowercase, hyphens only — rhdh-pr-review                           
  - description under 1024 chars — 491 chars, includes trigger phrases
  - description includes trigger phrases — "Use when...", "Also use when..."   
  - SKILL.md body under 500 lines — 78 lines                                   
  - Instructions use imperative form                                           
                                                                               
  Architecture                                                                 
                                                                               
  - N/A — no sub-commands, no setup gates, no register/mode system             
                                                                               
  References                                                                   
                                                                               
  - Domain knowledge split into references/ with clear pointers                
  - Each reference is self-contained — no transitive loading                   
  - Reference loading is conditional in the workflow (e.g., "Read              
  cluster-provisioning.md and follow the appropriate section")                 
  - No browser-only tools referenced                                           
                                                                               
  Scripts                                                   
                                                                               
  - N/A — no scripts in this skill                                             
  
  Quality                                                                      
                                                            
  - No time-sensitive URLs or dates                                            
  - Consistent terminology throughout
  - Concrete examples in references (PR #2756 example in operator-pr-images.md)
  - [!] One flag: VERSION=1.9 in cluster-provisioning.md will go stale — but we
   already added a comment per schultzp2020's review

durandom (commented):
## Review: Skill Architecture & Information Duplication

Nice work on the overall skill structure — the router pattern integration with the `rhdh` orchestrator is correct, the OLM vs non-OLM distinction is well-handled, and the workflow phases are logical. A few architectural concerns though:

### 1. `.claude/settings.json` should not be part of this PR

This PR adds a repo-wide `.claude/settings.json` that enables `rhdh@rhdh` for everyone who clones the repo. That's a separate policy decision, not related to this skill. Please remove it from this PR — it can be discussed and added independently if desired.

### 2. Don't duplicate repository knowledge — point to it

The skill's `references/cluster-provisioning.md` (122 lines) essentially re-documents how `rhdh-test-instance` works: its Prow CI, the `/test deploy operator` slash commands, `make install-operator`, `make deploy-operator`, etc. This information already lives in the [rhdh-test-instance README](https://github.com/redhat-developer/rhdh-test-instance) and its own docs.

**The principle for skills in this project:** A skill should teach the agent how to *interpret and apply* information, not *duplicate* it. If the information lives in a repository, the agent can read it from there. The skill's job is to know *where* to look and *how* to use what it finds.

Good example of what to do instead: The `rhdh-test-instance` README already documents the `/test deploy operator` slash commands. The skill should reference that README rather than re-explaining the commands inline. This way, when the upstream repo updates its commands or conventions, the skill doesn't become stale.

Concretely:
- `cluster-provisioning.md` should become a thin pointer: "Read the README at `redhat-developer/rhdh-test-instance` for provisioning instructions" + skill-specific interpretation guidance (e.g., "when reviewing an operator PR, use a 4h TTL because reviews are short-lived")
- Same applies to parts of `operator-pr-images.md` that describe CI workflow behavior already documented in `rhdh-operator`'s `.github/workflows/pr-container-build.yaml`

### 3. `rhdh-test-instance` is missing from the repo registry

The skill depends heavily on `rhdh-test-instance` but this repo isn't listed in `skills/rhdh/references/rhdh-repos.md` (which tracks all 12 RHDH ecosystem repos). It should be added there — with its key concepts (Prow CI, Makefile targets, slash commands). That way any skill that needs cluster provisioning can reference it centrally, rather than each skill re-documenting it.

Additionally, `cluster-provisioning.md` is not really skill-specific knowledge — it's ecosystem knowledge. Consider making it a shared reference under `skills/rhdh/references/` so other skills that need a cluster can reuse it.

### 4. Hardcoded paths in `cluster-provisioning.md`

The repo-discovery snippet has hardcoded, user-specific paths:

```bash
~/Documents/something-about-skills/rhdh-test-instance
~/src/rhdh/rhdh-test-instance
~/rhdh-test-instance
```

These are machine-specific guesses that won't work for other contributors. The `rhdh` CLI already handles repo discovery via `$RHDH` — the skill should use that mechanism instead of reimplementing it with hardcoded paths.

### 5. Missing `<cli_setup>` block

The workflow's `<tracking>` section uses `$RHDH log add` and `$RHDH todo add`, but the SKILL.md has no `<cli_setup>` block. Every other sub-skill that uses the CLI starts with one. Either add it or remove the `$RHDH` references from tracking.

---

**TL;DR:** The domain expertise in this skill is solid. The main ask is: don't duplicate what the repos already document — instead, teach the agent *where to look* and *how to interpret* what it finds there. That keeps the skill maintainable and prevents drift.

durandom (changes requested):

Fortune-Ndlovu:
done testing /unhold

skills/rhdh-pr-review/workflows/review-operator-pr.md
```diff
@@ -0,0 +1,443 @@
...
+Use `redhat-developer/rhdh-test-instance` — see `../../rhdh/references/rhdh-repos.md` for its capabilities, Makefile targets, and `/test deploy` slash commands. Read the repo's own README for full usage.
+
+- **No cluster at all** (`oc whoami` fails) → use the rhdh-test-instance PR workflow: comment `/test deploy operator <version> 4h` on a PR via `gh pr comment`. Use a 4h TTL (reviews are short-lived). Match the version to the PR's target branch.
+- **Cluster accessible but no RHDH** → use rhdh-test-instance locally: `make install-operator` then `make deploy-operator`. Read the repo's README for `.env` setup.
+
+Once the operator and Backstage CR are healthy, proceed to Phase 4.
+
+---
+
+## Phase 4: Swap Operator Image
```
rm3l: The PR may contain changes to the Operator default configuration. So simply swapping the image may not be enough. The Operator may actually fail to start/reconcile without the default configuration it expects.
- For OLM, I would suggest deploying the whole operator bundle image from the PR instead
- And for non-OLM, it would make more sense to swap the image in the (generated) `install.yaml` manifest from the PR branch and apply it. See https://github.com/redhat-developer/rhdh-operator/blob/main/dist/rhdh/install.yaml

Fortune-Ndlovu: That's a really important insight, swapping just the operator image assumes the binary is self-contained, but if the PR changes default config (like files under config/profile/ or default-config/), those configs are baked into the bundle/manifests, not the operator binary itself. The operator would start but could fail to reconcile because it expects config that isn't there. Ill update the skill, thanks

rm3l (approved):

durandom (approved):
All 5 points from my earlier review are addressed:

1. `.claude/settings.json` removed from PR
2. Duplication eliminated — skill now points to upstream repos instead of re-documenting
3. `rhdh-test-instance` added to `rhdh-repos.md`
4. Hardcoded paths removed
5. `<cli_setup>` block added

rm3l's catch on full bundle deployment (vs image-only swap) was the most important change — nicely handled.

Minor nit for a follow-up: the routing table in `skills/rhdh/SKILL.md` uses `@rhdh-pr-review` while other routes use file paths (`../overlay/SKILL.md`). The "To route" instruction below it is correct though, so not blocking.

LGTM, nice work Fortune!

PR #22, opened 2026-05-07
feat(overlay): add generate-metadata workflow for Package entity creation

davidfestal:
## Summary

- Add `workflows/generate-metadata.md` — a 6-phase workflow that generates missing Package metadata YAML files and audits existing metadata for consistency within overlay workspaces
- Add `scripts/derive-metadata.py` — a Python helper script for deterministic metadata derivation (name shortening, OCI URL, supportedVersions, plugins-list parsing, env var extraction)
- Add `<path_resolution>` and `<shell_permissions>` directives to `SKILL.md` for reliable script invocation and sandbox-safe GitHub API calls across all workflows
- Rewrite `references/metadata-format.md` with real-world examples, correct `catalog-entities/extensions/` paths, and comprehensive field documentation
- Delegate onboard-plugin Phase 4 to the new generate-metadata workflow
- Add `tests/unit/test_derive_metadata.py` with 47 tests covering all public functions

## Workflow Phases

1. **Workspace Identification** — resolve target workspace
2. **Scan for Missing Metadata** — detect plugins without metadata files, flag consistency issues
3. **Fetch Upstream Source** — retrieve package.json + config.d.ts via `gh api` (agent-direct, no subprocess)
4. **Analyze Config & Wiring** — generate appConfigExamples from config schemas, delegate frontend wiring
5. **Plugin Entity Resolution** — determine partOf references or Plugin entity creation
6. **Write Files & Report** — assemble YAML, update smoke-tests/test.env, audit existing metadata, propose commit/PR

## Test plan

- [x] 289 tests pass (47 new + 242 existing)
- [ ] Run generate-metadata workflow against a real overlay workspace (e.g., analytics) to validate end-to-end
- [ ] Verify onboard-plugin workflow still works with Phase 4 delegation

Made with [Cursor](https://cursor.com)

### DIFF
--- a/skills/overlay/SKILL.md
+++ b/skills/overlay/SKILL.md
@@ -1,7 +1,15 @@
 ---
 name: overlay
-description: Manages the rhdh-plugin-export-overlays repository — onboards plugins to the Extensions Catalog, updates plugin versions, fixes overlay build failures, triages and analyzes PRs, triggers publishes, and manages plugin workspaces. Use when working with overlays, importing plugins, debugging CI, checking PRs, or bumping versions.
+description: Manages the rhdh-plugin-export-overlays repository — onboards plugins to the Extensions Catalog, updates plugin versions, generates and audits Package metadata, fixes overlay build failures, triages and analyzes PRs, triggers publishes, and manages plugin workspaces. Use when working with overlays, importing plugins, generating metadata, auditing metadata, fixing metadata inconsistencies, debugging CI, checking PRs, or bumping versions.
 ---
+
+<path_resolution>
+All relative paths in this skill (`scripts/`, `references/`, `workflows/`, `templates/`) are relative to **this SKILL.md file's directory**. Derive the skill root from the absolute path used to read this file. For example, if this file was read from `/Users/me/.agents/skills/overlay/SKILL.md`, then `scripts/derive-metadata.py` resolves to `/Users/me/.agents/skills/overlay/scripts/derive-metadata.py`. Always use absolute paths when invoking scripts or reading reference files.
+</path_resolution>
+
+<shell_permissions>
+Prefer running `gh api` and `gh search code` as **direct shell commands** rather than via Python subprocess. Direct `gh` calls go through the user's command allowlist without triggering permission prompts. Python scripts that only do local work (file I/O, JSON processing, field derivation) also need no extra permissions. Only request `full_network` for Python scripts that internally spawn `gh` as a subprocess — the sandbox blocks network access from child processes.
+</shell_permissions>
 
 <cli_setup>
 This skill uses the orchestrator CLI. **Set up first:**
@@ -63,6 +71,7 @@ What overlay task would you like to do?
 2. **Update plugin version** — Bump to newer upstream commit/tag
 3. **Check plugin status** — Verify health and compatibility
 4. **Fix build failure** — Debug CI/publish issues
+8. **Generate or audit metadata** — Add missing Package metadata or fix inconsistencies in existing metadata
 
 ### Core Team Tasks
 
@@ -84,6 +93,7 @@ What overlay task would you like to do?
 | 2, "update", "bump", "upgrade", "version" | `workflows/update-plugin.md` |
 | 3, "status", "check", "health" | Run inline status checks |
 | 4, "fix", "debug", "failure", "error" | `workflows/fix-build.md` |
+| 8, "metadata", "generate metadata", "add metadata", "audit", "audit metadata", "fix metadata", "check metadata", "validate metadata" | `workflows/generate-metadata.md` |
 
 ### Core Team Routes
 
@@ -158,6 +168,7 @@ See `../rhdh/references/github-reference.md` for full patterns.
 | onboard-plugin.md | Full 6-phase process to add new plugin |
 | update-plugin.md | Bump to newer upstream version |
 | fix-build.md | Debug and resolve CI failures |
+| generate-metadata.md | Generate missing Package metadata and audit existing metadata for consistency |
 
 ### Core Team Workflows
 
--- a/skills/overlay/references/metadata-format.md
+++ b/skills/overlay/references/metadata-format.md
@@ -11,47 +11,139 @@ Two entity kinds work together:
 <package_entity>
 **Location:** `workspaces/<name>/metadata/<package>.yaml`
 
-**Purpose:** Defines a single exportable package with its configuration.
+**Purpose:** Defines a single exportable package with its OCI artifact reference and configuration.
+
+**Frontend plugin example:**
 
 ```yaml
 apiVersion: extensions.backstage.io/v1alpha1
 kind: Package
 metadata:
-  name: <package-name>  # e.g., backstage-plugin-aws-codebuild
-  title: <Display Name>
-  description: <Brief description>
+  name: backstage-community-plugin-dynatrace
+  namespace: rhdh
+  title: "Dynatrace"
+  links:
+    - url: https://red.ht/rhdh
+      title: Homepage
+    - url: https://github.com/backstage/community-plugins/issues
+      title: Bugs
+    - title: Source Code
+      url: https://github.com/backstage/community-plugins/tree/main/workspaces/dynatrace/plugins/dynatrace
+  annotations:
+    backstage.io/source-location: url:https://github.com/backstage/community-plugins/tree/main/workspaces/dynatrace/plugins/dynatrace
+  tags: []
 spec:
-  packageName: <npm-package-name>  # e.g., @aws/backstage-plugin-aws-codebuild
-
-  # Dynamic plugin configuration
-  dynamicPluginConfig:
-    frontend:
-      # or backend: for backend plugins
-      mountPoints:
-        - id: entity-card
-          importName: AwsCodeBuildCard
-          config:
-            layout:
-              gridColumnEnd: span 4
-
-  # Example app-config.yaml snippets
+  packageName: "@backstage-community/plugin-dynatrace"
+  dynamicArtifact: oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-community-plugin-dynatrace:bs_1.49.4__10.17.0!backstage-community-plugin-dynatrace
+  version: 10.17.0
+  backstage:
+    role: frontend-plugin
+    supportedVersions: 1.45.3
+  author: Dynatrace
+  support: community
+  lifecycle: active
+  partOf:
+    - dynatrace-community
   appConfigExamples:
-    - title: Basic Configuration
-      content: |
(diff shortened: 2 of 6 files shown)


skills/overlay/scripts/derive-metadata.py
```diff
@@ -0,0 +1,554 @@
...
+    ("parfuemerie-douglas", "parfdg"),
+]
+
+OCI_REGISTRY = "oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays"
+
+
+def shorten_name(name: str) -> str:
+    """Apply shortening rules only if name exceeds K8S_NAME_LIMIT."""
+    if len(name) <= K8S_NAME_LIMIT:
+        return name
```
durandom, line 48: If the shortening rules don't reduce the name below 63 chars, this returns a name that exceeds the Kubernetes limit. The upstream [`shorten-component-name.sh`](https://github.com/redhat-developer/rhdh-plugin-export-utils/blob/main/common/scripts/shorten-component-name.sh) likely has a final truncation step.

Add a safety net — e.g., truncate to 55 chars + `-` + 7-char hash suffix:

```python
if len(shortened) > K8S_NAME_LIMIT:
    import hashlib
    h = hashlib.sha256(name.encode()).hexdigest()[:7]
    shortened = shortened[:K8S_NAME_LIMIT - 8] + '-' + h
return shortened
```

The test `test_long_name_shortened` only passes because the specific test string happens to get short enough — it doesn't cover names where the rules aren't sufficient.

skills/overlay/scripts/derive-metadata.py
```diff
@@ -0,0 +1,554 @@
...
+    metadata_dir = workspace_dir / "metadata"
+    existing_files = list(metadata_dir.glob("*.yaml")) if metadata_dir.exists() else []
+    existing_names = {f.stem for f in existing_files}
+
+    missing = []
+    for plugin in plugins:
+        path = plugin["path"]
+        path_suffix = path.rstrip("/").split("/")[-1] if path != "." else ""
+        found = any(path_suffix and path_suffix in name for name in existing_names)
+        if not found and path != ".":
```
durandom, line 136: Substring matching: `"argocd" in "something-argocd-backend"` is `True`. If the workspace has both `plugins/argocd` and `plugins/argocd-backend`, and a metadata file `x-argocd-backend.yaml` exists, the check for `plugins/argocd` falsely matches.

Either:
- Use exact suffix matching with a separator: `name.endswith(path_suffix)` or `name == expected_metadata_name`
- Or better: derive the expected metadata name via `package_name_to_metadata_name` and check for exact file stem match

The Copilot comment (#4) flagged the same issue.

skills/overlay/scripts/derive-metadata.py
```diff
@@ -0,0 +1,554 @@
...
+    result.setdefault("support", "community")
+
+    if result["support"] in ("generally-available", "tech-preview"):
+        result["support_needs_confirmation"] = True
+
+    return result
+
+
+def run_gh(args, check=True):
+    """Run a gh CLI command and return stdout as string."""
```
durandom, line 270: This `run_gh` doesn't match the pattern from `analyze-pr.py` and `triage-prs.py`:

- Missing `FileNotFoundError` handler (crash if `gh` isn't installed)
- Returns raw stdout string instead of parsed JSON
- No `json.JSONDecodeError` handling

If you keep `run_gh` (see my comment about removing `fetch-and-derive`), align with the existing pattern.

skills/overlay/scripts/derive-metadata.py
```diff
@@ -0,0 +1,554 @@
...
+        ["api", f"repos/{owner_repo}/contents/{path}?ref={ref}", "--jq", ".[].name"],
+        check=False,
+    )
+    if not output:
+        return []
+    return output.splitlines()
+
+
+def fetch_and_derive_all(
+    workspace_dir: Path, workspace: str, source: dict, missing_paths: list[str]
```
durandom, line 311: This subcommand mixes network I/O (`gh api` via subprocess) with pure computation (field derivation). The workflow already describes the `gh api` calls as shell commands in Phase 3.1 — having the script *also* do them via subprocess is redundant.

This is also what creates the `<shell_permissions>` contradiction in SKILL.md: the directive says "prefer direct `gh` calls over Python subprocess" but this function does exactly the opposite.

Recommendation: remove `fetch-and-derive`. The workflow should call `gh api` directly (agent-friendly, no sandbox issues), then pipe the results to `derive` for each plugin. The `scan` + `derive` subcommands are sufficient.

skills/overlay/scripts/derive-metadata.py
```diff
@@ -0,0 +1,554 @@
...
+    env_cmd.add_argument("file", help="Path to YAML file")
+
+    args = parser.parse_args()
+
+    if not args.command:
+        parser.print_help()
+        sys.exit(1)
+
+    is_tty = os.isatty(sys.stdout.fileno())
+
```
durandom, line 432: `os.isatty(sys.stdout.fileno())` raises `io.UnsupportedOperation` in environments without a real fd (some test runners, CI wrappers). Use `sys.stdout.isatty()` instead.

Also: the existing scripts use an explicit `--json` flag rather than TTY auto-detection. That's more predictable — consider matching that pattern.

skills/overlay/references/metadata-format.md
```diff
@@ -11,61 +11,153 @@ Two entity kinds work together:
...
+  namespace: rhdh
+  title: "Dynatrace"
+  links:
+    - url: https://red.ht/rhdh
+      title: Homepage
+    - url: https://github.com/backstage/community-plugins/issues
+      title: Bugs
+    - title: Source Code
+      url: https://github.com/backstage/community-plugins/tree/main/workspaces/dynatrace/plugins/dynatrace
+  annotations:
```
durandom, line 32: `supportedVersions: 1.45.3` but the `dynamicArtifact` tag above uses `bs_1.49.4`. These should match — this is the reference doc that agents will follow when generating metadata.

Same issue with the backend example below (`supportedVersions: 1.48.3` vs `bs_1.49.4` in the tag).

skills/overlay/SKILL.md
```diff
@@ -1,8 +1,16 @@
...
 ---
 
+<path_resolution>
+All relative paths in this skill (`scripts/`, `references/`, `workflows/`, `templates/`) are relative to **this SKILL.md file's directory**. Derive the skill root from the absolute path used to read this file. For example, if this file was read from `/Users/me/.agents/skills/overlay/SKILL.md`, then `scripts/derive-metadata.py` resolves to `/Users/me/.agents/skills/overlay/scripts/derive-metadata.py`. Always use absolute paths when invoking scripts or reading reference files.
+</path_resolution>
+
+<shell_permissions>
+Prefer running `gh api` and `gh search code` as **direct shell commands** rather than via Python subprocess. Direct `gh` calls go through the user's command allowlist without triggering permission prompts. Python scripts that only do local work (file I/O, JSON processing, field derivation) also need no extra permissions. Only request `full_network` for Python scripts that internally spawn `gh` as a subprocess — the sandbox blocks network access from child processes.
+</shell_permissions>
```
durandom, line 12: This directive says "prefer direct `gh` shell commands over Python subprocess" but `derive-metadata.py`'s `fetch-and-derive` subcommand does exactly the opposite — it calls `gh api` via `subprocess.run`.

Either:
- Remove `fetch-and-derive` from the script (recommended — see my comment on the script), or
- Rewrite this to explain that the script handles `gh` calls internally and agents should use the script's subcommands rather than calling `gh` directly for metadata tasks

skills/overlay/SKILL.md
```diff
@@ -63,6 +71,7 @@ What overlay task would you like to do?
 2. **Update plugin version** — Bump to newer upstream commit/tag
 3. **Check plugin status** — Verify health and compatibility
 4. **Fix build failure** — Debug CI/publish issues
+8. **Generate or audit metadata** — Add missing Package metadata or fix inconsistencies in existing metadata
```
durandom, line 74: Nit: jumps from option 4 to option 8 in the user-facing menu. Should be 5 (or renumber the Core Team section to leave room).

durandom (changes requested):
## Review: Script Architecture & Quality

Nice work on the overall concept — extracting deterministic field derivation (name shortening, OCI URLs, version consistency) into a script is exactly the right pattern for this project. The workflow phases are logical and the delegation from `onboard-plugin` Phase 4 makes sense. The `metadata-format.md` rewrite with real-world examples is a big improvement.

However, the script has some architectural issues and quality gaps compared to the existing scripts (`analyze-pr.py`, `triage-prs.py`). See inline comments for specifics.

### Architectural

**1. `fetch-and-derive` mixes concerns — remove or split it**

The existing scripts in this project call `gh` to fetch data, but keep fetching and transformation cleanly separated. `fetch-and-derive` does both: it shells out to `gh api` via subprocess AND runs the derive logic. This creates the `<shell_permissions>` contradiction (see inline comment) and makes the subcommand harder to test (requires mocking subprocess).

The workflow already describes the `gh api` calls as direct shell commands in Phase 3.1. The script should only do local/deterministic work: `scan` (read local files) and `derive` (pure computation). Let the agent or the workflow handle the `gh api` calls — that's what the existing scripts' patterns do.

**2. `extract-env-vars` doesn't need to be a subcommand**

This is a one-liner: `grep -oP '\$\{[A-Z_][A-Z0-9_]*\}' file | sort -u`. Adding it as a Python subcommand adds code to maintain with no benefit.

### Functional Bugs

**3. `shorten_name` has no safety net** (see inline)

**4. `find_missing_metadata` uses substring matching** (see inline)

### Quality — Match Existing Patterns

**5. `run_gh` is inconsistent with existing scripts**

Both `analyze-pr.py` and `triage-prs.py` have `run_gh` that:
- Returns parsed JSON (not raw strings)
- Catches `FileNotFoundError` for missing `gh` CLI
- Catches `json.JSONDecodeError`

This script's `run_gh` does none of those. If you keep `run_gh`, follow the established pattern.

**6. Use `--json` flag instead of TTY auto-detection** (see inline)

### Minor

**7. Module docstring shows old usage** — The usage examples don't show subcommands (`scan`, `derive`, etc.) which is confusing.

**8. Numbering gap in intake menu** — Options jump from 4 to 8 (see inline).

**9. `supportedVersions` mismatch in reference examples** — Both examples in `metadata-format.md` show inconsistent versions (see inline).

---

**TL;DR:** The derive logic (name shortening, OCI URLs, consistency checks) is genuinely valuable as a script. The main asks: (1) remove `fetch-and-derive` — let the workflow handle `gh api` calls, keep the script doing only local/deterministic work, (2) fix the `shorten_name` safety net and `find_missing_metadata` matching, (3) align `run_gh` and output format with existing scripts.

PR #24, opened 2026-05-13
New Skill for RHDH Test Plan Review

psrna:
Reviews an RHDH test plan Jira ticket and suggests platform/integration version updates based on support lifecycle pages and RHDH release milestones. Use when given an RHDH test plan Jira ticket ID to check which platform/integration versions to add or remove. Use when asked to "update test plan", "review test plan", "check platform versions in test plan", "review RHDH test plan", "what platforms should we test for RHDH X", or "update supported versions in test plan".

### DIFF
--- a/skills/rhdh-test-plan-review/SKILL.md
+++ b/skills/rhdh-test-plan-review/SKILL.md
@@ -4,355 +4,75 @@ description: |
   Reviews an RHDH test plan Jira ticket and suggests platform/integration version updates based on support lifecycle pages and RHDH release milestones. Use when given an RHDH test plan Jira ticket ID to check which platform/integration versions to add or remove. Use when asked to "update test plan", "review test plan", "check platform versions in test plan", "review RHDH test plan", "what platforms should we test for RHDH X", or "update supported versions in test plan".
 ---
 
-# RHDH Test Plan Review
+<essential_principles>
 
-Reviews a test plan Jira ticket against platform and integration support lifecycle pages. Suggests adding versions GA before RHDH Code Freeze and removing versions EOL before RHDH GA date. No changes are applied without explicit user approval.
+<principle name="no_changes_without_approval">
+Never modify Jira tickets without explicit user approval. Collect all decisions first, summarize them, then ask whether to apply — direct update [d], comment [c], or discard [n].
+</principle>
 
-## Prerequisites
+<principle name="version_tracking_rules">
+Different platforms accumulate versions differently:
+- **OCP**: accumulate all active versions
+- **AKS, EKS, GKE, Quay**: single latest version only — replace, never accumulate
+- **ARO, OSD, ROSA**: single version each, evaluated independently — replace if a newer version is GA before code_freeze
+- **RHBK**: track major versions only (e.g., `26` not `26.0`); accumulate all active majors
+- **PostgreSQL**: RHDH support policy is the baseline; Backstage-only versions require a Jira Feature ticket warning
+</principle>
 
-**Jira (required):** Verify `acli` is configured:
+<principle name="milestone_cutoffs">
+Add/remove decisions are based on `code_freeze` and `ga_date` from the RHDH schedule sheet — not today's date.
+- **Add**: version GA date ≤ `code_freeze`
+- **Remove**: version EOL date ≤ `ga_date`
+</principle>
 
-```bash
-python ~/.claude/skills/rhdh-jira/scripts/setup.py --json
-```
+<principle name="adf_preservation">
+Jira descriptions are ADF (nested JSON). When updating, modify only version strings and date cells inside existing table cells — never convert to plain text and back. Preserve all other ADF structure exactly.
+</principle>
 
-If not configured, load `~/.claude/skills/rhdh-jira/SKILL.md` and follow its Prerequisites section before continuing.
+<principle name="token_safety">
+Never read `.jira-token` into context. Always use shell substitution: `"$(cat "$TOKEN_FILE")"`.
+</principle>
 
-**Google Sheets API (required):** Ensure `gcloud` is on your `PATH`, then check auth:
+</essential_principles>
 
-```bash
-python scripts/check_gsheets.py
-```
+<intake>
 
-If the check fails, load `references/google-sheets-setup.md` and walk the user through setup before continuing.
+## RHDH Test Plan Review
 
-## Workflow
+Provide a Jira ticket ID or URL (e.g., `RHIDP-8994`) to begin.
 
-### Step 1: Fetch the test plan Jira ticket
+**Wait for response before proceeding.**
 
-```bash
-acli jira workitem view TICKET-ID --json
-```
+</intake>
 
-Extract the RHDH version from `fixVersions[0].name`. If empty, check `summary` for a version string (e.g., "RHDH 1.6 Test Plan").
+<routing>
 
-Normalize the version: strip any non-numeric prefix ("RHDH ", "v", "rhdh-") to get plain `major.minor` (e.g., "1.6").
+| Input | Workflow |
+|-------|----------|
+| Jira ticket ID or URL | Read `workflows/review-test-plan.md` and follow it |
 
-If the version cannot be determined from either field, ask: "I couldn't determine the RHDH version from this ticket. What version is this test plan for?"
+</routing>
 
-### Step 2: Fetch RHDH milestone dates
+<reference_index>
 
-```bash
-python scripts/fetch_schedule.py --version "1.6"
-```
+| Reference | Purpose | Path |
+|-----------|---------|------|
+| sources | Lifecycle URLs and extraction guidance per platform/integration | `references/sources.md` |
+| google-sheets-setup | One-time gcloud auth setup for schedule sheet access | `references/google-sheets-setup.md` |
+| rhdh-jira auth | Jira REST API token setup and curl patterns | `~/.claude/skills/rhdh-jira/references/auth.md` |
 
-Use `--sheet-id <id>` only when the schedule lives in a different spreadsheet than the default (e.g. a fork or replacement sheet).
+</reference_index>
 
-Expected output:
-```json
-{
-  "version": "1.6",
-  "feature_freeze": "2025-09-15",
-  "code_freeze": "2025-10-01",
-  "ga_date": "2025-10-15",
-  "tab": "2025 Schedule"
-}
-```
+<success_criteria>
 
-If the script returns `{"error": "spreadsheet_not_found"}`, ask:
+- [ ] RHDH version extracted from ticket
+- [ ] Milestone dates fetched from schedule sheet
+- [ ] All platform and integration versions checked against lifecycle sources
+- [ ] Overview diff presented (key dates, platforms, integrations)
+- [ ] Each proposed change reviewed interactively (a/k/e)
+- [ ] User chose how to apply: [d] direct update, [c] comment, [n] discard
+- [ ] If [d]: Jira description updated via REST API; child tasks offered one at a time
+- [ ] If [c]: comment posted to ticket
+- [ ] If [n]: confirmed no changes were made
 
-> "I couldn't access the RHDH Release Schedule sheet (ID: `<spreadsheet_id>`). Please share the sheet URL or ID."
-
-Extract the spreadsheet ID from the URL — it is the long alphanumeric string between `/d/` and `/edit` (e.g., `https://docs.google.com/spreadsheets/d/<ID>/edit`). Then retry:
-
-```bash
-python scripts/fetch_schedule.py --version "1.6" --sheet-id <id>
-```
-
-If the script returns `{"error": "version_not_found"}`, ask: "I couldn't find RHDH [version] milestones in the schedule sheet. Could you confirm the exact version string as it appears in the sheet?"
-
-Use `code_freeze` and `ga_date` throughout the rest of the workflow.
-
-### Step 3: Parse the test plan description
-
-The description from `acli jira workitem view TICKET-ID --json` is in Atlassian Document Format (ADF). Parse the ADF JSON to locate:
-
-1. **Key dates table** — the first table in the description, with rows for Feature Freeze, Code Freeze, and GA. Each row has the milestone name in the first cell and the date in the second cell (may be empty).
-2. **Platform versions table** — rows listing OCP, ARO, AKS, EKS, GKE, OSD, ROSA with version numbers
-3. **Integration versions table** — rows listing PostgreSQL variants, RHBK, Quay with version numbers
-
-Record the current version set for each entry. Normalize version strings to `major.minor` for comparison.
-
-Also record the current key dates from the table (may be blank if not yet filled in).
-
-### Step 4: Fetch lifecycle data
-
-Load `references/sources.md` for all lifecycle URLs and how to extract version/date information from each source.
-
-For each source, fetch the page using WebFetch:
-- Retry up to 3 times on failure
-- If all 3 attempts fail: skip this source and record a warning — do not abort the entire run
-
-Apply these rules using `code_freeze` and `ga_date` from Step 2:
-- **Add**: version is GA on or before `code_freeze` AND is not already in the table
-- **Remove**: version reaches EOL on or before `ga_date` AND is currently in the table
-
-**Special rule — AKS, EKS, GKE (managed Kubernetes services):**
-These platforms test only the **latest available version**. Do not accumulate versions.
-- Identify the newest version that is GA on or before `code_freeze`
-- If it differs from the current version in the table, suggest **replacing** the current value with only that latest version
-- Do not suggest adding alongside the existing version
-
-**Special rule — ARO, OSD, ROSA:**
-Each of these platforms tracks a **single version** at a time (not necessarily the latest). Do not accumulate versions.
-- Suggest **replacing** the current version if a newer version is GA on or before `code_freeze` and not EOL before `ga_date`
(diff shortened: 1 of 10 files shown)


skills/rhdh-test-plan-review/references/sources.md
```diff
@@ -0,0 +1,96 @@
...
+**What to extract:** ROSA follows OCP version availability. Use OCP GA and EOL dates as a proxy. Apply the same add/remove rules as OCP.
+
+**Note:** The ROSA-specific lifecycle page (`docs.openshift.com/rosa`) returns 403 for automated access. The Red Hat lifecycle API has ROSA entries only up to 4.13 (classic architecture). OCP lifecycle data is the best available proxy for current versions.
+
+### ARO (Azure Red Hat OpenShift)
+
+**URL:** https://learn.microsoft.com/en-us/azure/openshift/support-lifecycle  
+**What to extract:** Kubernetes version GA dates and end of support (EOL) dates.
+
+### AKS (Azure Kubernetes Service)
```
zdrapela, line 31: Hi, I just finished a set of skills for lifecycle management of K8s jobs. I placed them in `openshift/release` repo, because that's where the jobs are defined, but now I'm not convinced it's the best place, because they cannot be used in combination with yours skill.

You can check them for inspiration: https://github.com/openshift/release/pull/79129. I chose a more programmatic approach, finding API endpoints and parsing JSONs.

skills/rhdh-test-plan-review/references/sources.md
```diff
@@ -0,0 +1,96 @@
+# Lifecycle Sources
+
+URLs and extraction guidance for each platform and integration. Load this file during Step 4 of the workflow.
+
+## Platforms
+
+### OCP (OpenShift Container Platform)
```
zdrapela, line 7: And I have similar for OCP: https://github.com/openshift/release/pull/78203 😅

skills/rhdh-test-plan-review/references/sources.md
psrna: My idea was mostly around keeping the Test Plan JIRA up to date for a given release and eventually having child tasks for suggested changes that are needed. As jira is the source of truth for our release execution, I wanted to have a skill that helps making sure we have valid coverage. But i see there is some overlap with yours. Is there any preference on what you would like to see should happen with the approaches?

skills/rhdh-test-plan-review/SKILL.md
```diff
@@ -0,0 +1,339 @@
...
+- The version strings within the platform and integration table cells in the ADF
+- The date cells in the key dates table (Feature Freeze, Code Freeze, GA rows — match by row label, update the date cell)
+
+Preserve all other ADF structure exactly — other tables, headings, paragraphs, and non-version rows must remain untouched.
+
+Update via REST API:
+
+```bash
+TOKEN_FILE="/opt/homebrew/bin/.jira-token"
+
```
durandom: Hardcoded macOS Homebrew path. The `rhdh-jira` skill discovers this dynamically:

```bash
ACLI_DIR=$(dirname "$(readlink -f "$(which acli)")")
TOKEN_FILE="$ACLI_DIR/.jira-token"
```

See `skills/rhdh-jira/references/auth.md` for the pattern. Same issue in the Gotchas section (line 337).

This breaks on Linux and non-Homebrew macOS installs.

skills/rhdh-test-plan-review/scripts/check_gsheets.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def colored(text, code):
+    if _is_tty:
+        return f"\033[{code}m{text}\033[0m"
+    return text
+
+
+def get_gcloud_token():
+    gcloud = shutil.which("gcloud")
```
durandom: This function is nearly identical to `fetch_schedule.py:get_gcloud_token()` — same fallback paths, same logic. Deduplicate: either extract to a shared module, or have `fetch_schedule.py` call `check_gsheets.py --json` as a pre-flight check.

Also: the fallback paths (`~/Downloads/google-cloud-sdk/`, `/opt/homebrew/bin/`) are macOS-specific. On Linux, gcloud is typically at `/usr/bin/gcloud`, `/usr/local/bin/gcloud`, or `/snap/bin/gcloud`. Since `shutil.which()` is checked first and handles the common case, consider just dropping the hardcoded fallbacks entirely — if gcloud isn't in PATH, the user should fix their PATH.

skills/rhdh-test-plan-review/scripts/fetch_schedule.py
```diff
@@ -0,0 +1,248 @@
...
+import argparse
+import json
+import os
+import re
+import sys
+from datetime import datetime
+from pathlib import Path
+
+SPREADSHEET_ID = "1knVzlMW0l0X4c7gkoiuaGql1zuFgEGwHHBsj-ygUTnc"
+
```
durandom, line 28: If this sheet is replaced or a new copy is made, the script breaks silently. Make it a CLI argument with this as the default:

```python
parser.add_argument(
    "--sheet-id",
    default="1knVzlMW0l0X4c7gkoiuaGql1zuFgEGwHHBsj-ygUTnc",
    help="Google Sheets spreadsheet ID",
)
```

The workflow in SKILL.md can still omit the flag (uses the default), but it becomes overridable.

skills/rhdh-test-plan-review/scripts/fetch_schedule.py
```diff
@@ -0,0 +1,248 @@
+#!/usr/bin/env -S uv run --script
+# /// script
```
durandom, line 2: This is a conscious exception to ADR-0002 (stdlib-only Python). The `uv run --script` pattern is a pragmatic choice — the Google Sheets SDK is the right tool here, and `uv` is already used in this project.

Worth adding a brief comment noting this is intentional:
```python
# Note: uses uv run --script for Google API deps (see ADR-0002 for stdlib-only rationale;
# Google Sheets SDK is an intentional exception)
```

skills/rhdh-test-plan-review/SKILL.md
```diff
@@ -0,0 +1,339 @@
+---
+name: rhdh-test-plan-review
+description: |
+  Reviews an RHDH test plan Jira ticket and suggests platform/integration version updates based on support lifecycle pages and RHDH release milestones. Use when given an RHDH test plan Jira ticket ID to check which platform/integration versions to add or remove. Use when asked to "update test plan", "review test plan", "check platform versions in test plan", "review RHDH test plan", "what platforms should we test for RHDH X", or "update supported versions in test plan".
+---
+
+# RHDH Test Plan Review
+
```
durandom, line 8: The workflow (Steps 1–8, ~300 lines) is inlined directly here. Other skills split this into `SKILL.md` (entry point with `<intake>`, `<routing>`, `<reference_index>`) and `workflows/<name>.md` (the actual process). This keeps SKILL.md scannable and the workflow separately readable.

Also missing the XML structure tags (`<essential_principles>`, `<intake>`, `<routing>`) that other skills use per the Agent Skills spec. See `skills/rhdh-pr-review/SKILL.md` or `skills/overlay/SKILL.md` for the pattern.

durandom (changes requested):
## Review: RHDH Test Plan Review Skill

Great concept — automating the test plan version review against lifecycle pages is exactly the kind of tedious, error-prone work that should be a skill. The workflow is thorough (lifecycle sources, special rules per platform, interactive review, child task creation), and the `fetch_schedule.py` milestone parser is well-structured.

A few things need fixing before merge.

### Blocking

**1. Hardcoded macOS paths — breaks on Linux** (see inline comments)

The Jira token path is hardcoded to `/opt/homebrew/bin/.jira-token`. The `rhdh-jira` skill discovers this dynamically — use the same pattern. Both Python scripts also have macOS-only gcloud fallback paths.

**2. Not wired into orchestrator or README**

Every other sub-skill is listed in `skills/rhdh/SKILL.md` (intake menu + routing table + skills_index) and `README.md`. Without this, the skill is undiscoverable. See PR #20 for an example of how to add the routing.

**3. No tests for `fetch_schedule.py`**

The script has several pure functions that are easy to test: `normalize_version`, `parse_date`, `find_milestones`, `find_schedule_tab`. Other scripts in this repo (`analyze-pr.py`, `triage-prs.py`) all have corresponding test files. The milestone-finding logic in particular (walking backwards from GA row, matching freeze keywords) is the kind of code that needs test coverage — it's parsing semi-structured sheet data with heuristics.

**4. `get_gcloud_token()` is duplicated**

Nearly identical function in both `check_gsheets.py` (line 22) and `fetch_schedule.py` (line 42), including the same hardcoded fallback paths. Extract to a shared module or have `fetch_schedule.py` call `check_gsheets.py --json` first and skip if it fails.

### Non-blocking

**5. `uv run --script` with Google API deps**

This is a conscious exception to ADR-0002 (stdlib-only Python). The `uv run --script` pattern is reasonable here — the Google Sheets SDK is the right tool for authenticated API access, and `uv` is already used in this project for `pytest`. Worth noting in a comment or the PR description that this is intentional, not accidental.

**6. SKILL.md structure doesn't follow the Agent Skills spec**

Other skills use `<essential_principles>`, `<intake>`, `<routing>`, `<reference_index>` XML tags. The workflow is inlined directly into SKILL.md (339 lines) rather than living in a separate `workflows/` file. Consider splitting to `workflows/review-test-plan.md` to match the project pattern and keep SKILL.md as the entry point.

**7. Hardcoded `SPREADSHEET_ID`**

If the sheet moves or is replaced, the script breaks silently. Make it a CLI argument with the current value as default.

**8. @zdrapela's comment is worth acknowledging**

They have overlapping lifecycle management work in `openshift/release` ([PR #79129](https://github.com/openshift/release/pull/79129), [PR #78203](https://github.com/openshift/release/pull/78203)) — using API endpoints and JSON parsing for the same platforms. Could be a collaboration opportunity or at least worth cross-referencing in `sources.md`.

---

**TL;DR:** The domain expertise and workflow design are solid. Main asks: (1) fix portability — no hardcoded macOS paths, (2) wire into orchestrator + README, (3) add tests for the schedule parser, (4) deduplicate `get_gcloud_token()`.

skills/rhdh-test-plan-review/references/sources.md
zdrapela: Your approach is fine. In my skills, I used more API endpoints and JSON parsing in bash scripts, which is more token-efficient. Maybe I can extract the reusable skills for programmatically retrieving supported versions into this repository, so they can be reused by your skill.

skills/rhdh-test-plan-review/SKILL.md
psrna: Hopefully resolved in recent commit. Thx!

skills/rhdh-test-plan-review/SKILL.md
psrna: I have used the skill-maker skill. If this is a conceptual thing or a pattern we want to follow, would it be possible to reflect/embed those principles in the skill-maker skill?

skills/rhdh-test-plan-review/scripts/fetch_schedule.py
psrna: Hopefully resolved in recent commit. Thx.

skills/rhdh-test-plan-review/scripts/check_gsheets.py
psrna: Hopefully fixed in recent commit. thx

skills/rhdh-test-plan-review/scripts/fetch_schedule.py
psrna: hopefully fixed too.

skills/rhdh-test-plan-review/SKILL.md
psrna: Added routing & intake

psrna: Hopefully resolved too. Following project patterns.

psrna:
@durandom I believe that I've resolved all your feedback above. Please take a final look and possibly merge. Thank you!

skills/rhdh-test-plan-review/scripts/fetch_schedule.py
```diff
@@ -0,0 +1,255 @@
+# Note: uses uv run --script for Google API deps (see ADR-0002 for stdlib-only rationale;
+# Google Sheets SDK is an intentional exception)
+#!/usr/bin/env -S uv run --script
```
gustavolira, line 1: The shebang needs to be the very first line of the file for the OS to recognize it when running `./fetch_schedule.py`. As written, line 1 is a comment, so `chmod +x && ./fetch_schedule.py` would either fail or fall through to `/bin/sh`. Move the note comment below the shebang:

```python
#!/usr/bin/env -S uv run --script
# Note: uses uv run --script for Google API deps (see ADR-0002 for stdlib-only rationale;
# Google Sheets SDK is an intentional exception)
```

skills/rhdh-test-plan-review/scripts/fetch_schedule.py
```diff
@@ -0,0 +1,255 @@
...
+        "code_freeze": ["code freeze", "code-freeze", "codefreeze"],
+        "feature_freeze": ["feature freeze", "feature-freeze", " ff "],
+    }
+
+    # Step 1: find the index of the GA row for this version
+    ga_index = None
+    for i, row in enumerate(rows):
+        cells = [str(c) for c in row]
+        row_text = " " + " ".join(cells).lower() + " "
+        version_match = ver in row_text or (version.lower().replace("rhdh", "").strip() in row_text)
```
gustavolira, line 160: This does plain substring matching -- `ver in row_text`. Searching for "1.1" would match a row containing "RHDH 1.10 GA" since "1.1" is a substring of "1.10". RHDH versions already cross this boundary. Should we use word-boundary regex instead?

```python
version_match = bool(re.search(rf'\b{re.escape(ver)}\b', row_text))
```

The second branch of the OR (`version.lower().replace("rhdh", "").strip()`) always produces the same result as `ver` after `normalize_version`, so it can be dropped at the same time.

skills/rhdh-test-plan-review/workflows/review-test-plan.md
```diff
@@ -0,0 +1,307 @@
...
+- **ADF round-trip**: Send ADF when updating via REST — converting to plain text destroys formatting. Modify only version strings inside existing table cells.
+- **Token safety**: Never read `.jira-token` into context. Use `"$(cat "$TOKEN_FILE")"` via shell substitution.
+- **Key dates table**: Match milestone rows by label keyword (e.g., "Code Freeze", "GA announce"). Update only the date cell. Leave rows you cannot match untouched.
+- **Version normalization**: Tables may use "v1.29", "1.29.x", or "Kubernetes 1.29" — normalize to `major.minor` before comparing.
+- **fixVersions format varies**: May be "1.6", "RHDH 1.6", "rhdh-1.6" — strip prefixes before passing to `fetch_schedule.py`.
+- **Schedule tab is year-based**: `fetch_schedule.py` tries current year first, then adjacent years. Pass `--sheet-id` if the schedule is in a non-default spreadsheet.
+- **Child task issuetype**: Use `"Task"` with a `parent` field. If that fails with 400, retry with `"issuetype": {"name": "Subtask"}`.
+- **Child task project key**: Use the same project key as the parent (e.g., `RHIDP`).
+
+</gotchas>
```
gustavolira, line 307: nit: Other workflows in the project (e.g., overlay workflows) include a `<success_criteria>` section, and `test_skill_structure.py::TestWorkflowStructure` validates this. The SKILL.md has success criteria, but the workflow file itself doesn't. Worth adding one here for consistency, even if the test currently only checks overlay workflows.

tests/unit/test_fetch_schedule.py
```diff
@@ -0,0 +1,199 @@
...
+        assert "ga_date" not in out
+        assert out["feature_freeze"] == "2025-05-20"
+        assert out["code_freeze"] == "2025-06-20"
+
+    def test_version_match_via_stripped_rhdh(self):
+        rows = [
+            ["Code freeze", "2025-01-05"],
+            ["RHDH 2.0", "GA ", "2025-02-01"],
+        ]
+        assert fs.find_milestones(rows, "RHDH 2.0")["code_freeze"] == "2025-01-05"
```
gustavolira, line 199: nit: No test for the version substring collision -- searching for "1.1" when the sheet contains "1.10". Given RHDH versions have already crossed the 1.9/1.10 boundary, this would protect a fix:

```python
def test_does_not_confuse_similar_versions(self):
    rows = [["RHDH 1.10", "GA date", "2025-08-01"]]
    assert fs.find_milestones(rows, "1.1") == {}
```

durandom:
## Re-review: blocking items addressed ✅

All four blocking items from my initial review are resolved — nice work, @psrna!

Two small things I noticed while re-checking. Non-blocking, up to you whether to fix now or in a follow-up:

### 1. `find_milestones` substring matching (from @gustavolira's review)

In `fetch_schedule.py`, `ver in row_text` means searching for version `"1.1"` would match rows containing `"1.10"`. RHDH has already crossed the 1.9→1.10 boundary, so this could bite in practice. A word-boundary check (e.g. `re.search(rf'\b{re.escape(ver)}\b', row_text)`) would fix it.

### 2. `<skills_index>` table in orchestrator

The routing works (intake option 9 + routing table are both there), but the `<skills_index>` summary table in `skills/rhdh/SKILL.md` still only lists the original 4 skills — `rhdh-test-plan-review` is missing from that table.

durandom (approved):
lgtm, up to @psrna to either fix here, follow up, or leave as is

PR #25, opened 2026-05-13
feat: extract CI lifecycle and Prow job management skills from openshift/release

zdrapela:
## Summary

- Extract 10 skills from `openshift/release` `.claude/skills/` into this repository
- Rename with concern-first naming (`lifecycle-*`, `prow-*`) and deduplicate shared scripts
- Add dual-mode support (local checkout or GitHub API via `gh` CLI) so skills work from any directory

### New skills

**Lifecycle (portable, pure API):** `lifecycle-ocp`, `lifecycle-aks`, `lifecycle-eks`, `lifecycle-gke`

**Prow job management (openshift/release):** `prow-ocp-jobs`, `prow-ocp-pools`, `prow-ocp-coverage`, `prow-aks-jobs`, `prow-eks-jobs`, `prow-gke-jobs`

**Shared infrastructure:** `_shared/resolve-repo.sh`, `_shared/fetch-yaml.sh`, `_shared/list-k8s-test-configs.sh`

### Key changes from originals

- 3x `list-k8s-test-configs.sh` duplication eliminated (was identical in AKS/EKS/GKE)
- 2x `print-configured-versions.sh` duplication eliminated (consolidated into `fetch-yaml.sh`)
- `ocp-lifecycle.jq` cross-skill reference is now a clean sibling path
- All read-only scripts auto-detect local `openshift/release` checkout or fall back to `gh` CLI
- `rhdh-decommission-release` intentionally left in `openshift/release` (destructive, local-only)


skills/_shared/rhdh_lifecycle/redhat.py
```diff
@@ -0,0 +1,205 @@
...
+    url = f"{LIFECYCLE_API_URL}?name={product_name.replace(' ', '+')}"
+    req = urllib.request.Request(
+        url, headers={"Accept": "application/json", "User-Agent": "rhdh-skill"}
+    )
+    try:
+        with urllib.request.urlopen(req, timeout=30) as resp:
+            return json.loads(resp.read().decode("utf-8"))
+    except (urllib.error.URLError, OSError) as exc:
+        print(f"ERROR: Failed to fetch lifecycle data for {product_name}: {exc}", file=sys.stderr)
+        sys.exit(1)
```
gustavolira: `fetch_api` is shared library code — `ocp.py`, `rhdh.py`, `check_ocp_lifecycle.py`, and `analyze_coverage.py` all import it. A `sys.exit(1)` here kills the caller's process on any network hiccup with no chance to handle it. The standalone scripts (`check_aks_lifecycle.py`, `check_eks_lifecycle.py`) already do this right — their local `fetch_json` returns `None` on failure. Should `fetch_api` do the same (return `None` or raise), and let the `if __name__` entry points decide whether to exit?

skills/_shared/rhdh_lifecycle/ocp.py
```diff
@@ -0,0 +1,102 @@
...
+Usage:
+    from rhdh_lifecycle.ocp import classify_ocp_versions
+    from rhdh_lifecycle.redhat import fetch_api
+    api_data = fetch_api("Red Hat OpenShift Container Platform")
+    versions = classify_ocp_versions(api_data, "2025-05-13")
+"""
+
+import re
+
+from rhdh_lifecycle.redhat import _is_date, _to_date
```
gustavolira: Importing private functions (`_is_date`, `_to_date`) across module boundaries breaks the convention that `_`-prefixed names are internal. If `redhat.py` renames or removes these, nothing warns that `ocp.py` broke — no public API contract was violated. Should we just make them public (`is_date`, `to_date`) since they're clearly needed outside `redhat.py`?

skills/_shared/rhdh_prow/repo.py
```diff
@@ -0,0 +1,63 @@
...
+    is_remote is True when no local checkout was found.
+    """
+    # 1. Explicit override
+    if explicit_dir is not None:
+        p = Path(explicit_dir)
+        if (p / _SENTINEL).is_dir():
+            return p.resolve(), False
+        print(
+            f"WARNING: explicit dir {explicit_dir} does not contain {_SENTINEL}",
+            file=__import__("sys").stderr,
```
gustavolira: nit: `file=__import__("sys").stderr` — `os` is already imported at the top of this file. Just add `import sys` and use `file=sys.stderr`. The `__import__` trick makes this harder to read for no benefit.

skills/lifecycle/scripts/rhdh_lifecycle/rhdh.py
```diff
@@ -0,0 +1,45 @@
...
+        v["maintenance_end"] = v.get("phases", {}).get("Maintenance support", "N/A")
+    return versions
+
+
+def fetch_lifecycle_api(product_name):
+    """Fetch raw API data. Delegates to redhat module."""
+    return fetch_api(product_name)
+
+
+def parse_rhdh_versions(api_data, filter_version=None):
```
gustavolira, line 34: This enrichment loop (lines 33-36) is identical to the one in `fetch_rhdh_lifecycle` above (lines 18-21) — same three keys, same `.get()` chains. Extract to a small helper like `_enrich(v)` and call it from both?

skills/lifecycle/scripts/check_aks_lifecycle.py
```diff
@@ -0,0 +1,165 @@
...
+sys.path.insert(0, str(Path(__file__).resolve().parent.parent.parent / "_shared"))
+from rhdh_prow.repo import resolve_repo_root
+from rhdh_prow.yaml import print_configured_versions
+
+AKS_API_URL = "https://releases.aks.azure.com/parsed_data.json"
+EOL_API_URL = "https://endoflife.date/api/azure-kubernetes-service.json"
+CONFIG_DIR = "ci-operator/config/redhat-developer/rhdh"
+
+
+def fetch_json(url):
```
gustavolira: This `fetch_json` is identical to `check_eks_lifecycle.py:46` and near-identical to `pg_lifecycle.py:28` (`_fetch_json`) — same urllib boilerplate, same timeout, same User-Agent, same error handling. Ironic given the PR description says "3x duplication eliminated." Should we add a shared `fetch_json` to `_shared/rhdh_lifecycle/` and import it?

skills/lifecycle/scripts/check_aks_lifecycle.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Deprecated version
+    try:
+        deprecated = regional[list(regional.keys())[0]][0]["Current"].get(
+            "DeprecatedVersion", "N/A"
+        )
+    except (KeyError, IndexError):
+        deprecated = "N/A"
+
+    # Cross-verify with endoflife.date
```
gustavolira, line 86: This entire EOL cross-verification block (lines 100-119) is copy-pasted from/to `check_eks_lifecycle.py` (lines 127-146) — same variables, same boolean logic for `eol`/`extendedSupport`, same sort. A shared `filter_supported_eol_entries(eol_data, today)` would deduplicate ~20 lines and keep the two scripts in sync when the logic changes.

skills/lifecycle/scripts/rhdh_lifecycle/pg.py
```diff
@@ -0,0 +1,110 @@
...
+import urllib.request
+
+PROVIDERS = {
+    "upstream": "https://endoflife.date/api/postgresql.json",
+    "rds": "https://endoflife.date/api/amazon-rds-postgresql.json",
+    "azure": "https://endoflife.date/api/azure-database-for-postgresql.json",
+}
+
+
+def _fetch_json(url):
```
gustavolira: nit: This file reimplements its own `_fetch_json` and all lifecycle logic from scratch, but it talks to endoflife.date — the same API that AKS/EKS scripts consume. Since this PR introduces `_shared/rhdh_lifecycle/` specifically to consolidate this kind of code, should `pg_lifecycle.py` use it too?

skills/lifecycle/scripts/check_aks_lifecycle.py
```diff
@@ -0,0 +1,165 @@
...
+                has_support = True
+            elif isinstance(eol, bool):
+                has_support = not eol
+            elif isinstance(eol, str) and eol > today:
+                has_support = True
+            if not has_support and isinstance(ext, str) and ext > today:
+                has_support = True
+            if has_support:
+                eol_supported.append(entry)
+        eol_supported.sort(key=lambda e: [int(x) for x in e["cycle"].split(".")], reverse=True)
```
gustavolira: nit: This inline `lambda e: [int(x) for x in e["cycle"].split(".")]` appears 9+ times across the PR (here, `check_eks_lifecycle.py:146`, `check_gke_lifecycle.py`, `analyze_coverage.py`, `list_cluster_pools.py`, etc.). `_ver_sort_key` already exists in `redhat.py:61` — should we make it public and reuse it everywhere?

skills/_shared/rhdh_prow/yaml.py
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/env -S uv run --script
```
gustavolira: nit: This file has a `#!/usr/bin/env -S uv run --script` shebang but no `if __name__ == "__main__"` block — it's purely a library module. The shebang suggests it's runnable, which is misleading. Remove it?

skills/_shared/rhdh_lifecycle/rhdh.py
```diff
@@ -0,0 +1,45 @@
...
+        v["full_support_end"] = v.get("phases", {}).get("Full support", "N/A")
+        v["maintenance_end"] = v.get("phases", {}).get("Maintenance support", "N/A")
+    return versions
+
+
+def rhdh_supported_ocp_versions(rhdh_data):
+    """Return sorted list of OCP versions supported by any active RHDH release."""
+    return sorted(
+        {ocp for v in rhdh_data if v["supported"] for ocp in v.get("ocp_versions", [])},
+        key=lambda x: [int(n) for n in x.split(".")],
```
gustavolira: nit: `_ver_sort_key` already exists in `redhat.py:61`, but this uses an inline `lambda x: [int(n) for n in x.split(".")]`. Minor, but since this is the shared package, it'd be nice to dogfood the shared helper.

zdrapela:
Thank you @gustavolira for the reveiw! I incorporated you suggestions. I'm not using Python that often and I appreciate them. 

I then thought about the skill structure and it looks like a better idea to move them into two bigger skills with dedicated workflows.

durandom (changes requested):
## Review Summary

Well-structured PR — the consolidated 2-skill architecture with router + workflow pattern is clean, the dependency graph is acyclic, and the subprocess boundary between prow and lifecycle data is the right call. Two bugs to fix before merge.

---

### 🐛 Bug: `check_aks_lifecycle.py` — JSON output corruption

When both `--test-pattern` and `--json` are passed, `print_configured_versions()` writes human-readable text to stdout **before** the JSON blob, corrupting the output. The EKS script correctly guards this:

```python
# check_eks_lifecycle.py (correct)
if args.test_pattern and not args.json_output:
    print_configured_versions(...)
```

The AKS script is missing the `not args.json_output` guard. This matters because `analyze_coverage.py` could call this script with `--json`.

**Fix:** Add `and not args.json_output` to the `if args.test_pattern:` condition in `check_aks_lifecycle.py`.

---

### 🔒 Security: `redhat.py:fetch_api` — Insufficient URL encoding

`product_name.replace(' ', '+')` is not proper URL encoding. Characters like `&`, `=`, `#`, `?` pass through unescaped into the query string. Currently mitigated by `PRODUCT_ALIASES`, but `resolve_product_name` falls through to the raw user string for unknown aliases — so `check_lifecycle.py --product "foo&bar=baz"` would inject into the URL.

**Fix:** Replace with `urllib.parse.quote_plus(product_name)`.

---

### 💡 Suggestion: Replace symlink with copied modules (Windows compatibility)

The symlink `prow/scripts/rhdh_lifecycle → ../../lifecycle/scripts/rhdh_lifecycle` breaks silently on Windows with `core.symlinks = false` (the symlink is checked out as a plain text file containing the path, and Python imports fail without error).

Only **2 facade files** in prow actually touch the symlink (`rhdh_prow/repo.py` and `rhdh_prow/yaml.py`), re-exporting ~180 lines of infrastructure code. The lifecycle *data* functions are already accessed via subprocess `--json` in `analyze_coverage.py`.

Proposed alternative: copy `repo.py` (63 lines) and `yaml.py` (116 lines) directly into `rhdh_prow/` as standalone modules. Delete the symlink. Each skill becomes truly self-contained, no cross-platform hazard, no `sys.path` manipulation needed.

---

### Minor (non-blocking, good for follow-up)

- **`ocp.py`**: `api_data.get("data", [{}])[0]` will `IndexError` if `data` is an empty list `[]`. Guard like `redhat.py` does.
- **`generate_test_entry.py`**: `--reference` format is not validated (only `--version` is). Malformed input crashes at `split(".")`.
- **`list_aks_jobs.py` / `list_eks_jobs.py` / `list_gke_jobs.py`**: `sys.argv` mutation is fragile — use `main(["--pattern", ...] + sys.argv[1:])` instead.
- **`check-redhat.md`**: Missing Output and Action sections. `check-ocp.md` and `check-rhdh.md` also lack Action sections.
- **Unresolved from prior review**: duplicated `fetch_json` boilerplate (3 copies), copy-pasted EOL cross-verification block (AKS↔EKS), inline version sort lambdas vs. existing `ver_sort_key`.

zdrapela:
Hi @durandom, I addressed your review and also the remaining from @gustavolira.

About _💡 Suggestion: Replace symlink with copied modules (Windows compatibility)_, I'm only worried about those files would go out of sync, like fixes or improvements in one wouldn't be updated in the second one. Would a line in `AGENTS.md` to keep them in sync do the trick?

durandom (approved):
All previous feedback thoroughly addressed — nice work on the consolidation.

**4 non-blocking items for follow-up:**

1. `analyze_coverage.py` fallback `_fetch_api` still uses `.replace(' ', '+')` — inconsistent with the `quote_plus` fix in `redhat.py`
2. `analyze_coverage.py` fallback parsers use `[{}][0]` which will `IndexError` on empty `data: []` — same guard as `ocp.py`/`redhat.py` should be applied
3. `generate_cluster_pool.py` is missing `--reference` format validation (was fixed in `generate_test_entry.py` but not here)
4. `ruamel.yaml` via PEP 723 inline metadata technically diverges from the stdlib-only rule — might warrant an ADR update or documented exception

PR #26, opened 2026-05-14
feat: add prow-trigger-nightly skill

zdrapela:
## Summary

- Ports the `trigger-nightly-job.sh` script from the rhdh repo to a stdlib-only Python skill
- Adds support for triggering nightly jobs from **both** `redhat-developer/rhdh` and `redhat-developer/rhdh-plugin-export-overlays` repos via the OpenShift CI Gangway REST API
- Since this now covers both repositories, it makes sense to extract the skill here rather than keeping it in the rhdh repo alone

## What's included

- `SKILL.md` with interactive workflow (list jobs, natural language selection, options, confirm, execute)
- `trigger_nightly_job.py` — Python 3.9+ stdlib-only replacement for the bash script
  - `--list`: fetches configured nightly jobs from both repos, prints a table
  - `--list-tags`: fetches available image tags from quay.io
  - `--job`: triggers a job with optional overrides
  - `--dry-run`: prints the equivalent curl command
- Flag validation per repo: overlay jobs accept fork overrides (`--org`, `--repo`, `--branch`) but error on image/alert flags (`--tag`, `--image-repo`, `--image-registry`, `--send-alerts`), matching the step-ref YAML definitions in openshift/release

### DIFF
--- a/skills/prow-trigger-nightly/SKILL.md
+++ b/skills/prow-trigger-nightly/SKILL.md
@@ -1,0 +1,152 @@
+---
+name: prow-trigger-nightly
+description: >-
+  Trigger RHDH nightly ProwJobs on demand via the OpenShift CI Gangway REST API.
+  Supports both rhdh and rhdh-plugin-export-overlays repos. Use when the user
+  wants to trigger, run, kick off, or start a nightly CI job, run an on-demand
+  E2E nightly test, list available nightly jobs, or trigger an overlay nightly.
+  Also use when the user mentions Gangway, "nightly job", "periodic-ci",
+  RC verification, testing a custom image, running CI against a fork, or
+  checking available image tags on quay.io.
+---
+# Trigger Nightly ProwJobs
+
+Trigger RHDH nightly ProwJobs via the OpenShift CI Gangway REST API.
+
+Supports two repositories:
+- **rhdh** — the main RHDH application (`periodic-ci-redhat-developer-rhdh-*-nightly`)
+- **rhdh-plugin-export-overlays** — plugin export overlays (`periodic-ci-redhat-developer-rhdh-plugin-export-overlays-*-nightly`)
+
+## Script Location
+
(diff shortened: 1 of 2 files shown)


durandom (approved):
## Review Summary

Clean, well-structured single-purpose CLI skill. Correct stdlib-only implementation, thorough error handling, good security practices (list-form subprocess, dedicated kubeconfig, proper URL encoding). The SKILL.md workflow with natural language mapping is excellent agent UX.

**Verdict: Approve** — no bugs, no security issues.

---

### Suggestions (non-blocking)

**1. Argument group label is misleading**

```python
overrides = parser.add_argument_group("RHDH job overrides (ignored for overlay jobs)")
```

These flags actually **error out** for overlay jobs (`sys.exit(1)`), they're not silently ignored. Consider: `"RHDH job overrides (not supported for overlay jobs)"`.

**2. Hardcoded `like:1.` tag filter will miss RHDH 2.x tags**

```python
f"?limit={limit}&onlyActiveTags=true&filter_tag_name=like:1."
```

When RHDH 2.0 ships, `--list-tags` will silently return no results. A broader numeric filter or a `--tag-prefix` arg would future-proof this.

**3. Redundant `os.environ["KUBECONFIG"]` assignment (nit)**

Every `oc` call already passes `--kubeconfig` explicitly via `oc_base`. The env var set in `main()` is redundant — not harmful, just unnecessary process-global mutation.

**4. No `--json` flag (consistency, non-blocking)**

The lifecycle scripts in PR #25 all support `--json` for structured output. Adding it to `--list` and `--list-tags` would enable programmatic consumption by other skills.

zdrapela:
Thank you for the review. Good points! I addressed them. The PR is ready for a merge now

durandom (approved):
All 4 suggestions addressed cleanly. The dual-prefix tag filter and --json flag are well-implemented additions. LGTM.

PR #27, opened 2026-05-19
fix: skill audit fixes — broken refs, descriptions, fallbacks

schultzp2020:
## Summary
- convert `create-plugin` and `rhdh-jira` to explicit XML router sections so intake, routing, and reference boundaries are clearer for agents; align `create-plugin` command metadata with that structure
- fix the audited skill follow-ups across `overlay`, `rhdh`, `rhdh-local`, `rhdh-pr-review`, and `rhdh-test-plan-review` by adding standalone fallbacks, correcting broken cross-skill references, expanding trigger phrases, and clarifying `gcloud_token.py` usage
- fix `skill-maker`'s own `<intake>` block to explicitly wait for user input before routing, and add a focused regression test so that pause instruction stays covered

## Test plan
- [x] `uv run pytest`


schultzp2020:
I audited all our skills against `skill-maker`. I'm eagerly merging this PR since it's a skill. If there are complaints we can revert.

PR #18, opened 2026-05-06
feat: add sub-command router and sprint ceremony workflows to rhdh-jira

schultzp2020:
## Summary

Adds a sub-command router and 5 workflow references to the rhdh-jira skill, covering all applicable Scrum ceremonies for the RHDH team. Also updates the skill-maker skill with patterns learned during development.

## rhdh-jira: New sub-commands

| Command | Ceremony | Description |
|---------|----------|-------------|
| `assign` | Any time | 5-layer deep assignee analysis: team roster (GraphQL `team.teamV2`), expertise profiling, sprint capacity, soft signals (bus factor, growth), context proximity |
| `refine` | Backlog Refinement | 6-check refinement: missing fields per exit criteria, duplicates, hierarchy integrity, unaddressed comments, staleness, sprint readiness |
| `plan` | Sprint Planning | Carryover report, velocity trend, per-member capacity, ready-for-planning queue, sprint fill suggestions, critical customer bugs, retro action items |
| `sprint-report` | Sprint Review/Demo | Committed vs completed, per-member breakdown, epic progress, demo checklist with naming conventions and venue routing |
| `release` | Feature Refinement + SOS | PI funnel states, feature matrix, epic roll-up, cross-team dependency map, blocker bugs, RN readiness, risk assessment. Intended to replace `rhdh-release-triage` |

## Architecture

- **Sub-command router** with `scripts/command-metadata.json` as single source of truth
- **API preference order**: acli → GraphQL → REST API (skip acli for bulk operations)
- **Cross-cutting conventions** centralized in SKILL.md (confirmation flow, error handling, team conventions)
- **Data flow documentation** between sub-commands (plan reuses assign capacity, sprint-report shares velocity pattern, release invokes assign)
- **Exit criteria single-sourced** in `references/workflows.md` — refine and release point to it

## Key GraphQL discoveries

- `team.teamV2(id, siteId)` returns full team roster directly — no REST Teams API needed
- `JiraComponentsField` fragment works: `components { edges { node { name } } }`
- `JiraTeamViewField.selectedTeam.fullTeam.members` returns roster via issue context
- No GraphQL mutation exists for issue assignment — acli/REST for writes
- `"Team[Team]" = {teamId}` works in JQL (corrected Gotcha 5)

## skill-maker updates

- **Deduplication check** with intentional duplication exception (avoid transitive loading)
- **Avoid hardcoded thresholds** writing pattern — describe the signal, not an arbitrary number
[...]

PR #6, opened 2026-03-13
feat: add local-testing skill and rhdh local CLI commands

durandom:
## Summary

Integrates `rhdh-local-setup` knowledge into the skill system, enabling agents to guide users through local RHDH plugin testing as part of the full plugin development lifecycle: **create → export → overlay → test locally**.

- **New `skills/rhdh-local/` skill** — intake menu routing to 4 workflows: enable-plugin, disable-plugin, switch-mode, test-plugin; plus references for customization system, env vars, troubleshooting
- **Self-contained `rhdh_local` Python package** (`skills/rhdh-local/rhdh_local/`) — 5 focused modules: `sync.py`, `compose.py`, `settings.py`, `health.py`, `backup.py`; standalone CLI entry point (`scripts/rhdh-local`) plus orchestrator delegation (`rhdh local`)
- **Trimmed `overlay/references/rhdh-local.md`** — focused on plugin testing patterns only (dynamic plugins config, test entities, Extensions Catalog, verification checklist); operational content moved to rhdh-local skill
- **`skills/rhdh/SKILL.md`** — menu item 11 + routing for "local testing" keywords
- **`rhdh local` CLI commands** — `status`, `apply`, `remove`, `up`, `down`, `health`, `backup`, `restore`, `plugins list`; auto-detects `rhdh-local-setup/` via 4-layer cascade (env → config → sibling-of-rhdh-local → skill-root)
- **Compose variant support** — `--ollama` flag selects Ollama as Lightspeed LLM provider; `--safety-guard` enables Llama Guard content filtering; provider-aware compose file matrix (4 combinations)
- **`cmd_doctor` extended** — checks for local_setup path, rhdh-local presence, sync status, port 7007
- **249 unit tests** (TDD per CLAUDE.md)

## Architecture

```
skills/rhdh-local/
├── rhdh_local/          # Self-contained Python package (stdlib only)
│   ├── __init__.py      # Public API re-exports
│   ├── cli.py           # Standalone CLI + cmd_* handlers
│   ├── sync.py          # Copy-sync operations (CUSTOMIZATION_FILES/GLOBS)
│   ├── compose.py       # Container runtime detection + compose builder
│   ├── settings.py      # LastRunSettings persistence (--last replay)
│   ├── health.py        # Health checks (runtime, port, containers, backend)
│   └── backup.py        # Backup/restore (tar.gz archives)
├── scripts/
│   ├── rhdh-local       # Standalone entry point
[...]

PR #31, opened 2026-05-21
Adjust jira skills for handling formatted descriptions

johnmcollier:
I was using the `rhdh-jira` skills to populate child epics for a Feature, and found that the formatted descriptions (e.g. `h1 Epic Goal`) were being formatted at all. I remember hitting something similar when interacting with the Jira MCP server.

In short, when interacting with issues via the Jira API, Jira expects issue descriptions to be formatted in the [Atlassian Document Format](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/) (ADF).

I debated how to best handle this, ADF is a verbose JSON-based format (see [here](https://gist.github.com/johnmcollier/a5ebbae65790486d78ec73792e52b99d) for an example), so maintaining it could be a pain. Claude came up with a Python script that can convert the jira markdown to ADF, but I'm undecided if this is the best approach, open to suggestions.

Otherwise, this PR makes a couple small changes based on how the `acli` parameters work for setting certain issue fields.


johnmcollier:
@schultzp2020 mind taking a look?

skills/rhdh-jira/references/to-issue.md
```diff
@@ -119,36 +119,48 @@ Run the pre-creation check from `references/duplicates.md`. Scope to the target
...
+
+Then set priority, component, and story points together in one REST call:
+
+```bash
+curl -s -X PUT -u "$AUTH" -H "Content-Type: application/json" \
+  -d '{
+    "fields": {
+      "priority": {"name": "Major"},
+      "components": [{"name": "Plugins"}],
+      "story_points": 5
```
schultzp2020: Hit this live — Jira returns 400 with "field cannot be set." For writes you want the custom field ID:

```json
"customfield_10028": 5
```

`rest-api-fallback.md` already has an example you can copy from.

skills/rhdh-jira/SKILL.md
```diff
@@ -150,7 +151,7 @@ Load only what the current task requires.
...
 5. **Team field has two JQL syntaxes.** `customfield_10001` cannot be used in JQL WHERE clauses. However, `"Team[Team]" = {teamId}` (using the team UUID, not display name) works. Use the UUID syntax for JQL filtering; use `customfield_10001.name` in post-processing only when you need the display name from JSON output.
-6. **ADF vs plain text.** Reading descriptions via `--json` returns Atlassian Document Format (nested JSON). Creating/editing with `--description` accepts plain text. Don't try to round-trip ADF through `--description`.
+6. **ADF is required for formatted descriptions.** Reading descriptions via `--json` returns Atlassian Document Format (nested JSON). Jira Cloud's editor is ADF-native — plain text and Jira wiki markup (`h1.`, `*bold*`) both render as literal characters in the UI. For formatted descriptions, build ADF JSON directly in Python and pass via `--description-file`. Both `acli create` and `acli edit` accept ADF JSON via `--description-file`. Do not use Jira wiki markup in description files expecting it to render.
```
schultzp2020: Gotcha #6 still tells agents to "build ADF JSON directly in Python," but this PR adds `jira-wiki-to-adf.py` for exactly that. I'd point the gotcha at the script so people use it instead of hand-rolling ADF each time.

Something like: fill the wiki template → run `jira-wiki-to-adf.py` → pass the JSON via `--description-file`.

skills/rhdh-jira/references/to-epic.md
```diff
@@ -86,27 +86,37 @@ Run the pre-creation check from `references/duplicates.md`. Search RHIDP Epics (
 
 ### Step 8 — Create Epic
 
-Fill the template. Create the issue:
+Fill the template. Then convert to ADF using the helper script (see Gotcha #6). `acli create` accepts ADF via `--description-file`:
```
schultzp2020, line 89: Same pattern shows up in `to-feature.md` and `to-issue.md`. `/tmp/` won't work on Windows — might be worth a relative path, `$(mktemp ...)`, or a note to use `tempfile` when scripting this.

skills/rhdh-jira/references/to-epic.md
```diff
@@ -86,27 +86,37 @@ Run the pre-creation check from `references/duplicates.md`. Search RHIDP Epics (
...
-  --assignee "ACCOUNT_ID" \
-  --priority "Major" \
-  --component "Plugins" \
-  --yes
+  --description-file /tmp/epic-desc.adf.json \
+  --assignee "ACCOUNT_ID"
 ```
 
-If a parent Feature exists, link via REST:
+Then set priority, components, size, and parent Feature link together in one REST call. Cross-project parent links use `customfield_10018`, not `parent` (see Gotcha #16):
```
schultzp2020: The issuelinks warning is spot on. I'd just soften the "not `parent`" part — both of these worked when I tried them:

- `"customfield_10018": "RHDHPLAN-XXX"`
- `{"parent": {"key": "RHDHPLAN-XXX"}}`

RHIDP-14210 comes back with `parent.key` set and `customfield_10018` null, so either write shape is fine on our instance. Maybe something like "use Parent Link / `parent.key`, not issuelinks" reads clearer.

skills/rhdh-jira/scripts/jira-wiki-to-adf.py
```diff
@@ -0,0 +1,200 @@
...
+
+    return {"version": 1, "type": "doc", "content": content}
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
+        print("Usage: jira-wiki-to-adf.py <input.txt> [output.json]", file=sys.stderr)
+        sys.exit(1)
+
+    with open(sys.argv[1]) as f:
```
schultzp2020: Not a blocker — our other skill scripts use `argparse` + `--help` (ADR 0002, same as `parse_issues.py` / `validate_components.py`). Easy win to match that here, and add `encoding="utf-8"` on the file I/O while you're at it.

If you have a minute, a small unit test against `assets/examples/epic-example.txt` (check for `heading`, `bulletList`, `taskList` blocks) would keep the converter from regressing silently.

schultzp2020 (changes requested):
Nice work on the ADF path — this should fix the formatted-description issue for epics and features.

One thing I noticed: `templates.md`, `acli-commands.md`, and `support.md` still show the old plain-text create flow, and they aren't in this diff. Depending on which file an agent loads, it might still try the broken path. If you have bandwidth, updating those here (or in a quick follow-up) would keep the skill consistent. Happy to pair on a `skill-maker` audit pass if that's useful.

Test issues RHIDP-14212–14216 are closed.

skills/rhdh-jira/references/to-issue.md
johnmcollier: Fixed — changed to `"customfield_10028": 5` to match the field ID from `rest-api-fallback.md`.

skills/rhdh-jira/SKILL.md
johnmcollier: Updated Gotcha #6 to: "fill a wiki markup template then run `scripts/jira-wiki-to-adf.py <input.txt> <output.json>` to convert to ADF". Points agents at the script instead of hand-rolling JSON.

skills/rhdh-jira/references/to-epic.md
johnmcollier: Updated all three files (`to-epic.md`, `to-feature.md`, `to-issue.md`). ADF output files now use `$(mktemp)` stored in a shell variable (e.g. `EPIC_ADF=$(mktemp)`), referenced consistently through the conversion and `--description-file` steps. Added an inline comment noting Windows users should use `%TEMP%` or Python `tempfile`.

skills/rhdh-jira/references/to-epic.md
johnmcollier: Softened. Changed to: "Cross-project parent links accept either `customfield_10018` or `parent.key` — do not use `issuelinks`". Keeps the key warning (not issuelinks) while reflecting that both write shapes work.

skills/rhdh-jira/scripts/jira-wiki-to-adf.py
johnmcollier: Done:
- Switched to `argparse` with `--help` (positional `input`, optional `output`)
- Added `encoding="utf-8"` on both read and write
- Removed the now-unused `import sys`
- Added `tests/unit/test_jira_wiki_to_adf.py` with 10 tests against `epic-example.txt`: checks `heading`, `bulletList`, `taskList` blocks, heading levels, task item states, bold inline marks, and valid JSON output. All pass.

johnmcollier:
Also addressed your general comment about the old plain-text create flow:

- `acli-commands.md`: updated the description write guidance to reference `scripts/jira-wiki-to-adf.py` and `--description-file` instead of plain text
- `templates.md`: updated the create example to show the wiki→ADF conversion step before `acli create`, and removed the erroneous `--yes` flag (Gotcha #18)

durandom (approved):
Clean, well-tested stdlib-only converter that solves a real problem. All prior review feedback addressed. LGTM.

**3 non-blocking suggestions for follow-up:**

1. Italic regex `_([^_\n]+)_` can match mid-word underscores (e.g. `field_name_here`) — worth a docstring note if extending the converter later
2. `test_cli_stdout` doesn't actually test the CLI subprocess — consider renaming to `test_json_round_trip`
3. Multi-level lists not supported — documented limitation, acceptable for current scope

PR #43, opened 2026-06-05
feat: add rhdh-release skill for release management (v0.5.0)

durandom:
## Summary

- Adds new `rhdh-release` skill for RHDH release management (dates, status, teams, freeze announcements, blockers, CVEs, release notes)
- **14-command CLI** (`scripts/release.py`) for deterministic data gathering — the agent calls the CLI first, then adds judgment
- **Zero cross-skill symlinks** — all dependencies are either local copies or discovered at runtime
- **`gog` CLI** for Google Sheets/Docs access — replaces `gcloud`/`uv run --script`/Google SDK with stdlib-only Python
- **4 scripts** total: `release.py` (CLI + business logic), `formatters.py`, `jql.py`, `slack_templates.py`
- 13 JQL templates, 4 Slack announcement templates, 13 workflows with CLI-first routing
- 54 unit tests + structural checks

### Architecture

```
User → SKILL.md (routing) → release.py (CLI, deterministic) → agent (judgment)
                                 ├── jql.py (13 templates from jql-release.md)
                                 ├── slack_templates.py (4 templates)
                                 ├── formatters.py (auto JSON/human output)
                                 └── gog CLI (Google Sheets/Docs access)
```

### Cross-skill dependency removal

| Before | After |
|--------|-------|
| `scripts/gcloud_token.py` → symlink to rhdh-test-plan-review | Deleted (replaced by `gog` CLI) |
| `scripts/check_gsheets.py` → symlink to rhdh-test-plan-review | Deleted (replaced by `gog sheets metadata`) |
| `scripts/formatters.py` → symlink to rhdh skill | Local copy (stdlib-only, changes rarely) |
| `scripts/fetch_team_mapping.py` (separate script) | Inlined into release.py |
| `scripts/schedule.py` (separate script) | Inlined into release.py |
| `references/release-process.md` (150-line stale copy) | `gog docs cat <doc-id>` — live Google Doc on demand |
| `PARSE_ISSUES` hardcoded `../../rhdh-jira/` path | `_find_parse_issues()` — checks `~/.claude/skills/rhdh-jira/` first, falls back to sibling |
| Google Sheets via `uv run --script` + google-api-python-client | `gog sheets get/metadata` — no pip deps |

## Demo session

All output captured live on 2026-06-15.

### 1. Prerequisites check

```
$ python3 scripts/release.py --human check
  ✓ acli: /usr/local/bin/acli
  ⚠ .jira-token: missing (optional — acli may authenticate via other methods)
[...]

PR #42, opened 2026-06-04
feat: add agent-ready skill (RHDH-aware, consolidated)

johnmcollier:
## Summary

- Adds **agent-ready** skill: runs the [agentready](https://github.com/ambient-code/agentready) CLI against a git repository, parses the assessment JSON, and walks through failing findings in tier order.
  - **Auto mode** — applies deterministic fixes immediately; prompts only when project-specific input is needed. Skips ADR/design intent findings to avoid fabricating rationale; notes them in the final summary.
  - **Review mode** — prompts yes/skip/defer/quit per finding; asks for human rationale before writing any ADR or design doc.
  - **RHDH-aware** — detects the repo from its git remote URL against `rhdh-repos.md`; uses matched context to pre-fill AGENTS.md sections and skip inapplicable findings. Degrades gracefully if the rhdh skill is not installed.
  - **Batch mode** — scans a local directory for RHDH repos and presents a score table; user picks any repo to drill into.
  - **Inline AGENTS.md generation** — no delegation; uses assessment data (detected languages, CI commands, repo context) to generate AGENTS.md and CLAUDE.md in one pass.
- Updates README with a new **Repository Readiness** section.

## Test plan

- [ ] Run `agent-ready` on a repo with a low agentready score; verify findings are presented in tier order
- [ ] Verify `agent_instructions` finding generates AGENTS.md inline (not delegated)
- [ ] Verify RHDH repo detection works when remote URL matches `rhdh-repos.md`; gracefully skips if rhdh skill not installed
- [ ] Run in auto mode: confirm ADR/design findings are skipped, noted in summary
- [ ] Run in review mode: confirm each finding prompts; confirm ADR finding asks for rationale before writing
- [ ] Run batch mode: confirm only matched RHDH repos are assessed

PR #45, opened 2026-06-08
feat: add NFS migration and Backstage upgrade skills

rohitkrai03:
## Fixes
- https://redhat.atlassian.net/browse/RHIDP-14477

## Summary

- **nfs-migration** — Guide and agent skill for migrating RHDH plugins from the legacy Backstage frontend system to the New Frontend System (NFS). Includes reference files for each extension type, before/after examples, gotchas, and verification checklists. All API patterns verified against the Backstage source code and validated by migrating the quay plugin end-to-end.
- **backstage-upgrade** — Agent skill for upgrading `@backstage/*` dependencies using `backstage-cli versions:bump`, with RHDH version alignment, changelog-based breaking change review, and moved package migration. This skill is also used by the NFS migration skill.

## Test plan

- [x] `pytest` — 283 tests pass
- [x] `ruff check` — clean
- [x] Migrated `@backstage-community/plugin-quay` using the skill — `tsc`, `build`, and all 79 tests pass
- [ ] Review guide and skill content for accuracy

### DIFF
--- a/docs/nfs-migration-guide.md
+++ b/docs/nfs-migration-guide.md
@@ -2,6 +2,8 @@
 
 A practical guide for Red Hat Developer Hub plugin authors migrating from the legacy Backstage frontend system to NFS.
 
+> **Agent skill users:** The `nfs-migration` skill (`skills/nfs-migration/`) contains the same patterns broken into reference files optimized for agent consumption. This guide is the authoritative human-readable source. When updating migration patterns, update this guide first, then sync the corresponding reference file.
+
 ---
 
 ## 1. What is the New Frontend System
--- a/docs/nfs-migration-guide.md
+++ b/docs/nfs-migration-guide.md
@@ -1,0 +1,482 @@
+# Migrating RHDH Plugins to the New Frontend System (NFS)
+
+A practical guide for Red Hat Developer Hub plugin authors migrating from the legacy Backstage frontend system to NFS.
+
+---
+
+## 1. What is the New Frontend System
+
+The Backstage New Frontend System (NFS) replaces the legacy frontend plugin API. Instead of manually wiring plugins into an app with `createPlugin`, `createRoutableExtension`, `FlatRoutes`, and imperative JSX route trees, NFS uses declarative extension **Blueprints** (`PageBlueprint`, `ApiBlueprint`, `EntityContentBlueprint`, etc.) and `createFrontendPlugin` from `@backstage/frontend-plugin-api`.
+
+The app assembles itself from features:
+
+```ts
+import { createApp } from '@backstage/frontend-defaults';
+
+const app = createApp({ features: [myPlugin, catalogPlugin, ...] });
+```
+
+Plugins declare what they provide. The app decides what to render.
+
+---
+
+## 2. Why Migrate
+
+- **Declarative**: plugins describe their own routes, nav items, and APIs -- no more manual wiring in the app
+- **Configurable**: extensions can be enabled, disabled, or reordered via `app-config.yaml`
+- **Auto-discoverable**: apps can detect installed plugins automatically
+- **Composable**: modules can inject extensions into other plugins (e.g. entity tabs into catalog)
+- **Required**: the legacy APIs are being deprecated and will be removed
+
+---
+
+## 3. Deprecation Timeline
+
+| Phase | What happens |
+|-------|-------------|
+| **Current (RHDH 1.10)** | NFS available as `/alpha` exports alongside legacy |
+| **Next release (GA)** | NFS becomes the root export (`.`); legacy moves to `/legacy` with `@deprecated` tags |
+| **GA + 2 releases** | Legacy `/legacy` exports removed entirely |
+
+---
+
+## 4. Key Concepts
+
+### Blueprints vs Legacy Extension Factories
+
+Blueprints are declarative factories that replace imperative helpers like `createRoutableExtension()`. Each blueprint type (`PageBlueprint`, `ApiBlueprint`, `EntityContentBlueprint`) knows how to register itself with the app. Nav items are auto-discovered from pages -- no separate blueprint needed.
+
+```ts
+// Legacy
+export const MyPage = myPlugin.provide(createRoutableExtension({ ... }));
+
+// NFS
+const myPage = PageBlueprint.make({ params: { path: '/my-plugin', loader: () => ... } });
+```
+
+### `createFrontendPlugin` vs `createPlugin`
+
+| Legacy `createPlugin` | NFS `createFrontendPlugin` |
+|---|---|
+| `id: 'my-plugin'` | `pluginId: 'my-plugin'` |
+| `apis: [createApiFactory(...)]` | APIs go in `extensions` array as `ApiBlueprint` |
+| Pages/routes wired externally | Pages declared as `PageBlueprint` in `extensions` |
+| Named export | **Default export** |
+
+### `createFrontendModule`
+
+Bundles extensions that target *another* plugin. Common cases:
+
+- **Translations** target `pluginId: 'app'`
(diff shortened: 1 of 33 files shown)


schultzp2020:
@rohitkrai03 I think this skill would be better served upstream rather than here. https://github.com/backstage/backstage/tree/master/docs/.well-known/skills

rohitkrai03:
> @rohitkrai03 I think this skill would be better served upstream rather than here. https://github.com/backstage/backstage/tree/master/docs/.well-known/skills

@schultzp2020 This skill also has knowledge about RHDH specific extensions and blueprints that were converted from old dynamic plugin mount points to NFS APIs. It was created based on the idea that we would need to provide some kind of a guide to customers that they can use to learn about NFS migration and a skill that can help with that migration.

I can create a generic one suited for upstream but I can see its already there - https://github.com/backstage/backstage/blob/master/docs/.well-known/skills/plugin-new-frontend-system-support/SKILL.md.

schultzp2020:
> > @rohitkrai03 I think this skill would be better served upstream rather than here. https://github.com/backstage/backstage/tree/master/docs/.well-known/skills
> 
> @schultzp2020 This skill also has knowledge about RHDH specific extensions and blueprints that were converted from old dynamic plugin mount points to NFS APIs. It was created based on the idea that we would need to provide some kind of a guide to customers that they can use to learn about NFS migration and a skill that can help with that migration.
> 
> I can create a generic one suited for upstream but I can see its already there - https://github.com/backstage/backstage/blob/master/docs/.well-known/skills/plugin-new-frontend-system-support/SKILL.md.

Can you create a PR to update the generic one with the new findings you have here?

rohitkrai03:
> > > @rohitkrai03 I think this skill would be better served upstream rather than here. https://github.com/backstage/backstage/tree/master/docs/.well-known/skills
> > 
> > 
> > @schultzp2020 This skill also has knowledge about RHDH specific extensions and blueprints that were converted from old dynamic plugin mount points to NFS APIs. It was created based on the idea that we would need to provide some kind of a guide to customers that they can use to learn about NFS migration and a skill that can help with that migration.
> > I can create a generic one suited for upstream but I can see its already there - https://github.com/backstage/backstage/blob/master/docs/.well-known/skills/plugin-new-frontend-system-support/SKILL.md.
> 
> Can you create a PR to update the generic one with the new findings you have here?

@schultzp2020 You mean including RHDH specific references?

schultzp2020:
> > > > @rohitkrai03 I think this skill would be better served upstream rather than here. https://github.com/backstage/backstage/tree/master/docs/.well-known/skills
> > > 
> > > 
> > > @schultzp2020 This skill also has knowledge about RHDH specific extensions and blueprints that were converted from old dynamic plugin mount points to NFS APIs. It was created based on the idea that we would need to provide some kind of a guide to customers that they can use to learn about NFS migration and a skill that can help with that migration.
> > > I can create a generic one suited for upstream but I can see its already there - https://github.com/backstage/backstage/blob/master/docs/.well-known/skills/plugin-new-frontend-system-support/SKILL.md.
> > 
> > Can you create a PR to update the generic one with the new findings you have here?
> 
> @schultzp2020 You mean including RHDH specific references?

No, just the Backstage specific ones.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "rhdh-skill"
-version = "0.3.2"
+version = "0.4.0"
```
durandom: **Blocker: version downgrade.** Main is currently at `0.5.0` (from the rhdh-release skill merged in PR #43). This PR sets it to `0.4.0` — a downgrade.

Fix: rebase on current main and bump to `0.6.0` (minor bump, since these are new skills). Remember to update all 4 occurrences across 3 files:
- `pyproject.toml`
- `.claude-plugin/plugin.json`
- `.claude-plugin/marketplace.json` (2 occurrences)

skills/rhdh/SKILL.md
```diff
@@ -180,6 +180,14 @@ What would you like to do?
 
 **To route:** Read `../rhdh-test-plan-review/SKILL.md` and follow its intake process.
 
+### Backstage Upgrade Routes
+
+| Response | Skill |
+|----------|-------|
+| "upgrade backstage", "bump backstage", "update @backstage", "backstage version", "align deps", "versions:bump" | Route to `@backstage-upgrade` skill |
```
durandom: Nit: `@backstage-upgrade` uses an `@` prefix. No other skill in the routing tables uses this convention — they all use plain names (e.g. `rhdh-local`, `rhdh-pr-review`). Should be `backstage-upgrade` for consistency.

skills/backstage-upgrade/references/determine-target.md
```diff
@@ -0,0 +1,36 @@
+# Determine Target Backstage Version
+
+## For RHDH plugins
+
+RHDH pins a specific Backstage version per release. Load the version matrix from `../../rhdh/references/versions.md` to find the mapping.
+
+Ask the user: **"Which RHDH version are you targeting?"**
```
durandom, line 9: Non-blocking: this relative path (`../../rhdh/references/versions.md`) assumes both skills are installed side-by-side. If `backstage-upgrade` is installed individually via `npx skills add`, this path won't resolve. Same issue flagged on PR #42 for the `agent-ready` skill.

Suggestion: add a note in the skill description or a gotchas section that RHDH version alignment requires the `rhdh` core skill to be installed alongside. Graceful degradation ("ask the user for the target version directly") is already implied by the fallback to "specific version" — just make the dependency explicit.

docs/nfs-migration-guide.md
```diff
@@ -0,0 +1,482 @@
...
+
+const myDrawer = AppDrawerContentBlueprint.make({
+  params: {
+    title: 'My Drawer',
+    loader: () => import('./components/MyDrawer').then(m => <m.MyDrawer />),
+  },
+});
+```
+
+**Global header menu items** -- `GlobalHeaderMenuItemBlueprint`:
```
durandom, line 257: Good callout that `NavItemBlueprint` was removed. This is a common trap for people following older migration guides. Worth noting that this also affects code in `references/api-changes.md` — keeping both in sync is the maintenance concern mentioned in the top-level review.

durandom (changes requested):
## Review: NFS Migration & Backstage Upgrade Skills

High-quality skill content — the NFS migration guide (`docs/nfs-migration-guide.md`) is genuinely excellent. Real PR references, gotcha lists, before/after examples, RHDH-specific extensions (drawer, header, homepage). This is the kind of knowledge that's hard to find in one place.

### What's good

- **Comprehensive reference material** — 12 reference files, 3 before/after examples, real migration PRs from rhdh-plugins. The mount-point mapping table alone saves hours of digging.
- **Composable design** — `backstage-upgrade` can be called standalone or chained from `nfs-migration`. The cross-skill references are clean.
- **Follows existing skill patterns** — intake → routing → workflow → references structure matches other skills in the repo.
- **Validated against a real plugin** — the quay migration is a strong signal that the patterns work end-to-end.

### Blocking

1. **Version bump goes backward** — see inline comment on `pyproject.toml`

### Non-blocking suggestions

2. **No tests** — the PR description claims "283 tests pass" but adds zero test files for either skill. Existing skills in this repo have structural tests under `tests/unit/`. Since these are pure markdown skills with no scripts, this may be intentional — but worth clarifying.
3. **NFS migration guide duplicates skill reference content** — `docs/nfs-migration-guide.md` (482 lines) overlaps significantly with the skill's reference files (api-changes.md, gotchas.md, migrate-page.md). This creates a maintenance burden — changes to one must be reflected in the other. Consider making the guide the authoritative human-readable source and having the skill reference files link to it (or vice versa), rather than duplicating content.
4. See inline comments for minor consistency issues.

PR #73, opened 2026-07-22
feat(test-placement): add an authoring step and close the L1/L2 gap

gustavolira:
## The problem

The skill could route a developer correctly to *"L2, in `rhdh`"* — and then have nothing to say.

Steps 2–3 conclude with a **layer**, but Step 4 was organised by **repo**, and its `rhdh`
section covered only L3, L4a and L4b. L1 and L2 existed nowhere except as two cells in the
Step 3 ladder table. Across the whole skill, `mockServices`, `backend-test-utils` and
`frontend-test-utils` appeared **zero** times.

So the decision output and the creation guidance did not line up. I suspect that mismatch
is *why* L1/L2 fell through — it follows from the organising axis, not from an oversight.

## What changed

**Step 4** — the `rhdh` section is reordered cheapest-first to match the Step 3 ladder
(L1 → L2 → L3 → L4a → L4b) and gains the missing L1 and L2 entries. Each names a template
that exists on `main` today:

| Layer | Template |
|---|---|
| L1 | `plugins/scalprum-backend/src/service/router.test.ts` |
| L2 | `plugins/scalprum-backend/src/service/router.integration.test.ts` |

**Step 5 (new)** — turns the skill from an advisor into an author. It is a *procedure*,
not a knowledge dump: open the template Step 4 named, consult `rhdh: docs/testing.md`,
mirror the imports and setup, run the layer's command, apply a per-layer checklist.

## Why the API details are not in the skill

Deliberate. The Backstage test surface moves — `@backstage/test-utils` was renamed to
`@backstage/frontend-test-utils`, and the package went 0.5.1 → 0.6.1 inside 123 commits of
`main`. Nobody bumps a skill when a package is renamed, so an API name pinned here goes
stale silently and starts telling people to import something that no longer exists.

A path to a file on `main` cannot go stale the same way, and Step 5 carries a glob fallback
for the case where the template itself is moved. Same rule the skill already follows for
`layer-migration-matrix.md`: **the decision lives in the skill, the detailed reference lives
in the repo.**

This depends on [rhdh#5140](https://github.com/redhat-developer/rhdh/pull/5140)
(RHIDP-13234), merged, which added `docs/testing.md` — the L1–L3 authoring reference Step 5
points at. Added to References.

## One opinionated line

Step 5 states that a test must be justified by the failure it catches, not by a coverage
delta — Codecov is `informational: true` on both the project and patch statuses in `rhdh`
and `rhdh-plugins`, so no coverage number can block a PR. Worth encoding, since a PR
justified that way has already been rejected in review.

## Also

The frontmatter description gains *"how do I write a test for this"* / *"write a test for X"*
triggers. The previous list was all **where**-questions, which would not fire for someone
asking the skill to actually write the test.

## Verification

- All eight `rhdh` paths cited in the diff verified against `main` via the GitHub API.
- Frontmatter is valid YAML; `name` matches the directory and the description is 638 chars
  (`tests/unit/test_skill_structure.py` requires `20 < len <= 1024`).
- Diff is +39 / −7 on one file; the skill is 152 lines.

Per `AGENTS.md` this is a behaviour change, so it wants a **patch** tag after merge.

### DIFF
--- a/skills/test-placement/SKILL.md
+++ b/skills/test-placement/SKILL.md
@@ -4,10 +4,12 @@ description: >-
   Propose where to test a change in the RHDH dynamic-plugin ecosystem: which
   repo (rhdh-plugins, rhdh-plugin-export-overlays, rhdh), which test layer
   (unit, integration, component, cluster-free E2E, cluster E2E), where the
-  test lives, and how to create it. Use when asked "where should I test
-  this", "does this need a cluster", "should this be an e2e test", "which
-  repo does this test belong in", "what test layer", "test placement", or
-  when reviewing a test added at the wrong layer.
+  test lives, and how to create it — then write it by mirroring an existing
+  test in that repo. Use when asked "where should I test this", "does this
+  need a cluster", "should this be an e2e test", "which repo does this test
+  belong in", "what test layer", "test placement", "how do I write a test
+  for this", "write a test for X", "add a unit/integration/component test",
+  or when reviewing a test added at the wrong layer.
 ---
 
 # Test Placement Advisor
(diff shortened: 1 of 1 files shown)


schultzp2020 (approved):
/lgtm

PR #82, opened 2026-08-06
feat(skills): add rhdh-bump-yarn for multi-repo Yarn Berry bumps

nickboldt:
## Summary
- Add `rhdh-bump-yarn` skill + `scripts/bump-yarn.js` to bump Yarn Berry (binaries, yarnPath/packageManager/Containerfile pins) across rhdh-plugins, rhdh, overlays, rhdh downstream, and rhdh-plugin-catalog
- Default refreshes `yarn.lock` via `yarn install --mode=skip-build` (opt out with `--no-refresh-locks`)
- Document the skill in README under a new Yarn section

## Test plan
- [ ] `node skills/rhdh-bump-yarn/scripts/bump-yarn.js --help`
- [ ] `--scan --root <repo>` on at least one checkout
- [ ] `--dry-run --to 4.17.1 --root <repo>` shows expected binaries/files
- [ ] Optional: real bump on a throwaway clone with/without `--no-refresh-locks`

Generated-by: cursor

Ref: https://redhat.atlassian.net/browse/RHIDP-16074

### DIFF
--- a/skills/rhdh-bump-yarn/scripts/bump-yarn.js
+++ b/skills/rhdh-bump-yarn/scripts/bump-yarn.js
@@ -7,9 +7,13 @@
     - replace matching .yarn/releases/yarn-<from>.cjs binaries
     - rewrite yarnPath / packageManager / "yarn set version" / ENV YARN= pins
 
-  By default, refreshes yarn.lock via `yarn install --mode=skip-build` in
-  touched workspaces (matches Renovate lock metadata bumps). A full
-  five-repo lock regen can take >45 minutes. Use --no-refresh-locks to skip.
+  By default, refreshes yarn.lock via `yarn install --mode=update-lockfile`
+  for every lock that will run under --to yarn — including nested workspaces
+  that inherit a root yarnPath / packageManager (not only dirs whose pins
+  were rewritten). Skips locks with an explicit pin outside --from/--to
+  (e.g. roadie 4.9.2, backstage 4.8.1, dcm 4.15.0) and dist-dynamic
+  artifacts. New yarn-*.cjs binaries are chmod +x (100755). A full five-repo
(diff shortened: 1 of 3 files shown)


skills/rhdh-bump-yarn/scripts/bump-yarn.js
jonkoops, line 1: Pretty sure that most of this can be handled by [running two commands](https://yarnpkg.com/getting-started/install#updating-yarn):

```shell
yarn set version stable
yarn install
```

Basically, this just needs to run everywhere there is a `package.json` with a `packageManager` field with Yarn configured. Any agent should be clever enough to find those using some CLI-foo. No need to carry a bunch of custom JS code for it.

nickboldt: Good call — for normal workspaces `yarn set version <exact>` + lock refresh is the right API; we shouldn’t reimplement Berry binary download / `packageManager`+`yarnPath` rewrites.

Slimmed the skill/script accordingly: matching `packageManager` dirs now run `yarn set version <to>` (exact version so we stay aligned with the Renovate/reference bump, not `stable`). What’s left in the JS is only what Yarn can’t see across this five-repo set:

- selective `--from` (leave roadie/backstage/dcm / Yarn 3 alone)
- Containerfile / `ENV YARN=` / embedded `yarn set version` text
- orphan `.yarn/releases` (distgit) without `packageManager`
- Fullsend helper hardcodes (rhdh-plugins)
- inherited `yarn.lock` refresh (YN0028 after root-only bumps)

Happy to trim further if you think any of those extras belong in the skill doc only (agent CLI-foo) rather than the script.

---
Decisions / wording here: **Auto** (Cursor agent router), powered by **Composer**.

jonkoops (approved):

PR #86, opened 2026-08-12
feat(skills): add rhdh-must-gather-helm-bump skill and script

nickboldt:
## Summary

Adds the `rhdh-must-gather-helm-bump` skill (`skills/ci/`) and bundled `scripts/bump-must-gather-helm.py` to propagate a **Helm CLI** version bump across:

| Repo | What the skill changes |
|------|------------------------|
| [`redhat-developer/rhdh-must-gather`](https://github.com/redhat-developer/rhdh-must-gather) | `Makefile` `HELM_VERSION`, CGW `artifacts.lock.yaml` **or** `vendor/helm/`, Stage 2a/2b on `Containerfile` and `.rhdh/docker/Containerfile` |
| [`gitlab.cee.redhat.com/rhidp/rhdh`](https://gitlab.cee.redhat.com/rhidp/rhdh) | Mirror helm-related files into `distgit/containers/rhdh-must-gather/`; regenerate the hermetic distgit `Containerfile`; patch must-gather Tekton prefetch; write `sync/upstream_SHA_rhdh-must-gather` |

This is the Helm **CLI** in the must-gather image, not the RHDH Helm chart (`oci://quay.io/rhdh/chart`). The script does **not** commit, push, or open a PR/MR.

Catalog: registered under `skills/ci/` in `catalog.json`, marketplace, and `/ask-rhdh` routes.

https://redhat.atlassian.net/browse/RHIDP-16046

## What the script does

**Preview (ungated):** `--check` probes CGW and prints `helm_version=` / `mode=cgw|vendor`. `--dry-run` prints the planned writes without touching trees. Approval to preview is not approval to apply.

**Apply** (after `/mutation-gate`; upstream and distgit/Tekton as separate operations):

1. **Choose install path** by probing upstream `hack/check-helm-binary-available.sh` (same probe for `--check` and apply, including `--skip-upstream`):
   - **CGW (preferred):** `hack/update-helm-lockfile.sh` → Konflux `generic` prefetch on distgit root → Stage **2a** active
   - **Vendor fallback:** `hack/update-vendor.sh helm` → `gomod` prefetch on `vendor/helm` → Stage **2b** active
2. **Flip Stage 2a/2b bidirectionally** (doc-only comments stay commented) on upstream `Containerfile`, `.rhdh/docker/Containerfile`, and the regenerated distgit `Containerfile`.
3. **Sync into distgit:** `Makefile`, `artifacts.lock.yaml`, entire `hack/`, `.rhdh/docker/Containerfile`, entire `vendor/` (on CGW, omit/delete `vendor/helm`; keep `websocat` etc.). Does **not** copy upstream root `Containerfile` onto hermetic distgit.
4. **Regenerate distgit `Containerfile`** from `.rhdh/docker/Containerfile`, preserving `RHDH_MUST_GATHER_VERSION` and the Brew/Konflux `ENV SUMMARY=` / `LABEL` footer. Increments `release="N"` → `N+1` and the `konflux.additional-tags` `version-release` suffix with a numeric boundary (`2.0-1` → `2.0-2`, `2.0-10` → `2.0-11` not `2.0-20`).
5. **Patch Tekton prefetch** for must-gather only: `.tekton/rhdh-must-gather-2-{pull,push}.yaml` and `must-gather.prefetch_input` in `.tekton-templates/components.yaml` (`generic` vs `gomod`). Other `components.yaml` entries are left untouched.
6. **Write `sync/upstream_SHA_rhdh-must-gather`** from a **committed** upstream HEAD. Refuses to pin a dirty HEAD (`--allow-dirty` still does not write a SHA from uncommitted work).

`--skip-upstream` syncs distgit + Tekton + SHA only. `--skip-downstream` bumps upstream only.

## Discovery and guards

- `--parent-dir` finds must-gather (`1-must-gather`, …) and midstream (`4-rhdh`, `rhdh-downstream`, `rhidp-rhdh`) only when the candidate has `distgit/containers/rhdh-must-gather` (skips a GitHub `rhdh` clone).
- Requires `hack/check-helm-binary-available.sh` on upstream; missing `bash` / `rsync` is a hard error.
- `git status` failures (`fatal: not a git repository`, etc.) die instead of looking clean.
- Requires Python 3.9+, `bash`, `curl`, `git`, `rsync`.

## Agent workflow (in `SKILL.md`)

1. Confirm `--to` against Helm releases.
2. Resolve checkouts.
3. `--check` + `--dry-run` (preview).
4. Apply only after `/mutation-gate`.
5. Review `git diff` + [references/verification.md](https://github.com/redhat-developer/rhdh-skills/blob/feat/RHIDP-16046-must-gather-helm-bump/skills/ci/rhdh-must-gather-helm-bump/references/verification.md) (unit tests, Konflux, E2E triage including Helm 4 OCI stdout).
6. Commit / PR or MR only when asked, again through `/mutation-gate` (Jira via `/rhdh-jira-link`).

References: install-paths (CGW vs vendor, tarball layout), tekton-prefetch, verification, helm4-notes.

Related upstream work: [rhdh-must-gather#284](https://github.com/redhat-developer/rhdh-must-gather/pull/284) (CGW binary), [rhdh-must-gather#282](https://github.com/redhat-developer/rhdh-must-gather/pull/282) (vendored fallback).

## Tests

`tests/unit/test_must_gather_helm_bump.py` (invoked via `sys.executable`):

- Real `--to` bump off the `4.2.3` fixture (lockfile/vendor stubs update `HELM_VERSION`)
- CGW sync omits `vendor/helm`, keeps websocat, scopes prefetch to must-gather, updates **pull and push** PLRs
- Vendor mode flips 2a→2b; a second case flips **2b back to 2a**
- `--dry-run` writes nothing
- `release="10"` / `2.0-10` → `11` / `2.0-11`
- Failed `git status`, dirty-HEAD SHA refusal, `rhdh-downstream` discovery, missing CGW probe script
- `--skip-upstream --check` still reports `mode=cgw` when upstream CGW is available (stale distgit `vendor/helm` does not win)

## Test plan

- [x] `uv run pytest tests/unit/test_must_gather_helm_bump.py`
- [x] `uv run pytest tests/unit/test_skill_catalog.py`
- [ ] `--check` / `--dry-run` smoke test against local must-gather + midstream checkouts

Generated-by: cursor


Fortune-Ndlovu (approved):
LGTM

skills/ci/rhdh-must-gather-helm-bump/SKILL.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Python 3.9+, `curl`, `git`, `rsync`
+- Network access to CGW mirror and both repo checkouts
+- Clean git trees in upstream and downstream (or pass `--allow-dirty`)
+- Upstream checkout on a branch that contains `hack/update-helm-lockfile.sh`
+
+## Usage
+
+```bash
+SKILL=skills/ci/rhdh-must-gather-helm-bump   # under rhdh-skill checkout
```
schultzp2020: The category folder is gone after install, so an agent that copies this Usage block misses the script and starts editing Tekton by hand. `rhdh-yarn-bump` uses a placeholder for that reason.

```suggestion
SKILL=<this skill's directory>
```

skills/ci/rhdh-must-gather-helm-bump/SKILL.md
```diff
@@ -0,0 +1,103 @@
...
+## Workflow
+
+1. Confirm target `--to` version against [Helm releases](https://github.com/helm/helm/releases).
+2. Resolve `--upstream` / `--downstream` (or `--parent-dir`).
+3. Run `--check`; read `mode=cgw` or `mode=vendor`.
+4. Run `--dry-run`, then the script without `--dry-run`.
+5. Review `git diff` in **both** repos.
+6. Run every verification gate — read [references/verification.md](references/verification.md).
+7. Commit / PR·MR only when the user asks.
+
```
schultzp2020: `--check` and `--dry-run` can stay ungated. The next line writes two other working trees. `/mutation-gate` calls that a write. `rhdh-base-images` gates the apply; `rhdh-prow-trigger` says a dry-run is not approval to execute. Completion below currently treats the real run as done before anyone signed off.

```suggestion
3. Run `--check` and `--dry-run`. That pair is the preview.
4. Apply only after `/mutation-gate` approval, with upstream and distgit/Tekton as separate operations. Approval to dry-run is not approval to write.
5. Review `git diff` in both repos, then the gates in [references/verification.md](references/verification.md).
6. Commit / PR or MR only when the user asks, again through `/mutation-gate`.
```

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
```diff
@@ -0,0 +1,689 @@
...
+def infer_mode(
+    *,
+    upstream: Path | None,
+    downstream: Path | None,
+    to_version: str,
+    skip_upstream: bool,
+    check_only: bool,
+) -> str:
+    mode = "cgw"
+    if upstream is not None and (not skip_upstream or check_only):
```
schultzp2020: If upstream is the sync source, probe it on apply the same way you do for `--check`. `--skip-upstream` currently looks at distgit: leftover `vendor/helm` (the test fixture always has this) plus an old lock for a different version becomes `mode=vendor`, while `--check` still prints `mode=cgw`. Then you copy a CGW lockfile and set gomod prefetch.

```suggestion
    if upstream is not None:
```

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
```diff
@@ -0,0 +1,689 @@
...
+def assert_clean_tree(path: Path, label: str, *, allow_dirty: bool) -> None:
+    if allow_dirty:
+        return
+    result = subprocess.run(
+        ["git", "-C", str(path), "status", "--porcelain"],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if result.stdout.strip():
```
schultzp2020: `check=False` plus "empty stdout means clean" treats `fatal: not a git repository` as success. That path is rc 128 with empty stdout, then distgit still gets written. Tests always pass `--allow-dirty`, so this never runs.

```suggestion
    result = subprocess.run(
        ["git", "-C", str(path), "status", "--porcelain"],
        capture_output=True,
        text=True,
        check=False,
    )
    if result.returncode != 0:
        die(
            f"{label} is not a git repository or git status failed ({path}): "
            f"{result.stderr.strip() or 'unknown error'}"
        )
    if result.stdout.strip():
```

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
```diff
@@ -0,0 +1,689 @@
...
+
+    footer = footer.replace(f'release="{current}"', f'release="{next_n}"')
+    if version:
+        footer = footer.replace(f"{version}-{current}", f"{version}-{next_n}")
+    else:
+        footer = re.sub(
+            rf'(konflux\.additional-tags="[^\"]*-){re.escape(current)}',
+            rf"\g<1>{next_n}",
+            footer,
+        )
```
schultzp2020, line 346: `replace("2.0-1", "2.0-2")` also rewrites `2.0-10`. Same hole in the no-version regex. Add a test with `release="10"` / `2.0-10`; the fixture only covers 1 -> 2.

```suggestion
    footer = footer.replace(f'release="{current}"', f'release="{next_n}"')
    if version:
        footer = re.sub(
            rf"{re.escape(version)}-{re.escape(current)}(?![0-9])",
            f"{version}-{next_n}",
            footer,
        )
    else:
        footer = re.sub(
            rf'(konflux\.additional-tags="[^\"]*-){re.escape(current)}(?![0-9])',
            rf"\g<1>{next_n}",
            footer,
        )
```

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
```diff
@@ -0,0 +1,689 @@
...
+                "1-must-gather",
+                "1-rhdh-must-gather",
+                "rhdh-must-gather",
+                "must-gather",
+            )
+            if found is None:
+                die(f"Could not find must-gather under {parent}")
+            upstream = found
+        if downstream is None:
+            found = discover_repo(parent, "4-rhdh", "rhdh", "rhidp-rhdh")
```
schultzp2020: `--parent-dir` never looks for `rhdh-downstream`, which is the name `/rhdh-context` uses. `rhdh` is the GitHub clone. Validate fails (no silent write), but a sibling GitLab checkout is never tried. Stronger still: only accept a candidate that has `distgit/containers/rhdh-must-gather`.

```suggestion
            found = discover_repo(parent, "4-rhdh", "rhdh-downstream", "rhidp-rhdh")
```

schultzp2020 (dismissed):
The CGW vs vendor split, and editing only must-gather's prefetch, are the right idea. I'd tighten the apply path before this runs against `rhidp/rhdh`. `--check` and `--skip-upstream` can pick different modes, `2.0-10` becomes `2.0-20`, and a failed `git status` looks clean. Tests don't hit those cases. On Windows the new file fails on shebang exec (`WinError 193`) before rsync is involved.

Comments below are the hunks I'd change. While you're there: don't write `upstream_SHA` from a dirty HEAD; require `hack/check-helm-binary-available.sh` and turn a missing `bash`/`rsync` into `die()`; run tests via `sys.executable` and actually bump `--to` off the 4.2.3 fixture (the lockfile stubs are no-ops); cover `--dry-run`, push.yaml, and 2b back to 2a. Same `skills/ci/...` path is in the script docstring. PR body still names `bump-must-gather-helm.sh`.

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
```diff
@@ -0,0 +1,720 @@
...
+        assert_clean_tree(downstream, "Downstream", allow_dirty=args.allow_dirty)
+        sync_to_distgit(upstream, downstream, mode, dry_run=dry_run)
+        regenerate_distgit_containerfile(downstream / DISTGIT_REL, dry_run=dry_run)
+        flip_helm_stages(
+            downstream / DISTGIT_REL / "Containerfile",
+            mode,
+            dry_run=dry_run,
+        )
+        update_tekton_prefetch(downstream, mode, dry_run=dry_run)
+        update_upstream_sha(
```
schultzp2020, line 721: This call runs after the distgit rsync and the Tekton rewrite. The script samples `started_dirty` before the upstream bump. A clean one-shot apply therefore pins the parent commit, but distgit already has the uncommitted bump. With `--allow-dirty`, or after an uncommitted bump in the two-step flow, the die here fires after those writes. Distgit is already mutated and the SHA file is missing.

`test_dirty_head_refuses_upstream_sha` only checks that the SHA file is absent. Die on a dirty HEAD before any downstream write. Pin only a commit that contains the synced bump.

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
```diff
@@ -0,0 +1,720 @@
...
+    script = upstream / "hack" / "check-helm-binary-available.sh"
+    if not script.is_file():
+        die(f"Missing hack/check-helm-binary-available.sh in {upstream}")
+    bash = require_on_path("bash")
+    result = subprocess.run(
+        [bash, str(script), version],
+        cwd=str(upstream),
+        check=False,
+    )
+    return result.returncode == 0
```
schultzp2020: Upstream `hack/check-helm-binary-available.sh` exits 2 when curl is missing or argv is wrong. It exits 1 when CGW has no tarballs. This treats every non-zero as vendor. The CLI never requires curl on PATH, even though SKILL.md lists it.

A missing curl or a failed fetch then copies `vendor/helm` and writes gomod prefetch onto a CGW tree. Die on exit 2 and on any unexpected code. Only exit 1 should select vendor.

skills/ci/rhdh-must-gather-helm-bump/SKILL.md
```diff
@@ -0,0 +1,108 @@
...
+python3 "${SKILL}/scripts/bump-must-gather-helm.py" --to 4.3.0 --check --parent-dir ~/RHDH
+
+# Dry-run, then apply (apply only after /mutation-gate)
+python3 "${SKILL}/scripts/bump-must-gather-helm.py" --to 4.3.0 --dry-run --parent-dir ~/RHDH
+python3 "${SKILL}/scripts/bump-must-gather-helm.py" --to 4.3.0 --parent-dir ~/RHDH
+```
+
+Full flag reference: `python3 "${SKILL}/scripts/bump-must-gather-helm.py" --help`.
+
+**Default:** updates upstream, syncs helm-related paths into distgit, flips Stage 2a/2b, regenerates distgit `Containerfile`, patches Tekton prefetch, writes `sync/upstream_SHA_rhdh-must-gather`. Does **not** commit, push, or open PR/MR.
```
schultzp2020: The workflow asks for upstream and distgit as separate operations. This default is one command that also writes the SHA. Agents will copy it. The script never commits, so a one-shot apply cannot pin a SHA that is the bump. Show `--skip-downstream`, commit, then `--skip-upstream` here until that is true. The description also offers prefetch diagnosis, but the steps below are apply-only, so that utterance still starts a bump.

schultzp2020 (dismissed):
Thanks. I would wait to merge until you fix the two apply-path comments.

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
nickboldt: Fixed in a2876ba: distgit/Tekton/SHA writes now die while upstream HEAD is dirty, including after a one-shot bump. Pin SHA only from `--skip-upstream` after that commit.

skills/ci/rhdh-must-gather-helm-bump/scripts/bump-must-gather-helm.py
nickboldt: Fixed in a2876ba: probe exit 1 still selects vendor; exit 2 and any other code die. curl is required on PATH.

skills/ci/rhdh-must-gather-helm-bump/SKILL.md
nickboldt: Fixed in a2876ba: Usage now shows `--skip-downstream`, commit, then `--skip-upstream`. Dropped the prefetch-diagnosis trigger from the description so that utterance does not start a bump.

skills/ci/rhdh-must-gather-helm-bump/references/helm4-notes.md
```diff
@@ -0,0 +1,23 @@
...
+```text
+Pulled: quay.io/rhdh/chart:2.0-59-CI
+Digest: sha256:...
+---
+apiVersion: v1
+```
+
+Piping `helm template … | kubectl apply` fails validation (`apiVersion not set, kind not set`) unless those status lines are stripped. Upstream E2E uses `helm_template_yaml()` in `tests/e2e/lib/test-utils.sh`.
+
+Must-gather **collection** scripts do not pipe template output to kubectl today. If a future collector adds that pattern, filter pull lines or use `--output-dir` + `kubectl apply -f`.
```
rm3l, line 18: I don't think this will ever happen. The must-gather collection scripts never create resources in the cluster, but only collect data. So seeing a `kubectl apply` in any of the collectors would be a red flag.
Right now, `kubectl apply -f` is indeed run as part of the E2E setup phase to deploy the necessary resources.

skills/ci/rhdh-must-gather-helm-bump/SKILL.md
rm3l, line 1: This skill may not be needed anymore. With the rewrite of the must-gather in Golang ([RHIDP-16540](https://redhat.atlassian.net/browse/RHIDP-16540)), it no longer depends on the Helm binary directly, but instead consumes the Helm Go SDK.

nickboldt: so we don't have to update any of these files with every new RHDH .y or .z update?

https://github.com/redhat-developer/rhdh-skills/pull/86/changes#diff-6899cb90b4882e4bbbaef3325a0e53d616bd6a59426376acecf5a079b63c6fb0R15-R23


| Repo | Path | What changes |
|------|------|--------------|
| [`redhat-developer/rhdh-must-gather`](https://github.com/redhat-developer/rhdh-must-gather) | repo root | `Makefile` `HELM_VERSION`, `artifacts.lock.yaml` **or** `vendor/helm/`, Stage 2a/2b, hack scripts |
| [`gitlab.cee.redhat.com/rhidp/rhdh`](https://gitlab.cee.redhat.com/rhidp/rhdh) | `distgit/containers/rhdh-must-gather/` | Mirror helm-related files; regenerate hermetic `Containerfile` from `.rhdh/docker/Containerfile` |
| Same (midstream) | `.tekton/rhdh-must-gather-2-{pull,push}.yaml` | `prefetch-input` (`generic` vs `gomod`) |
| Same | `.tekton-templates/components.yaml` | `must-gather.prefetch_input` only |
| Same | `sync/upstream_SHA_rhdh-must-gather` | Upstream commit SHA after sync |

rm3l: Correct. That was indeed part of the simplification we wanted to achieve with the full rewrite to Go.

PR #90, opened 2026-08-17
feat(skills): add refinement skills

yangcao77:
Adds a seventh editorial category, `refinement/`, holding the OpenSpec artifact-driven change workflow: twelve model-invoked skills that create, implement, audit, verify, and archive a change under `openspec/changes/<name>/`, plus two reference skills (`rhdh-spec-driven-schema` for the artifact schema and templates, `openspec-journal` for the append-only event log) they drive through.

### What's here

- `openspec-new-change`, `openspec-continue-change`, `openspec-propose`, `openspec-ff-change` — create a change and its artifacts (`proposal.md`, `specs/`, `design.md`, `tasks.md`), differing only in pace: stop after the folder, one artifact at a time, or everything in one pass.
- `openspec-explore` — a thinking-partner mode for investigating problems before or during a change, usable at any point.
- `openspec-apply-change` — implements pending tasks against the artifacts, logging progress through `openspec-journal`.
- `openspec-audit-change` (pre-implementation, cross-artifact consistency) and `openspec-verify-change` (post-implementation, code-vs-artifact) close a change from opposite ends.
- `openspec-sync-specs`, `openspec-archive-change`, `openspec-bulk-archive-change` — merge delta specs into `openspec/specs/` and archive one or many finished changes.
- `openspec-onboard` — a narrated, real-work walkthrough of the full cycle for first-time users.

`docs/refinement/refinement-workflow.md` (with a diagram) explains the real artifact dependency graph — `specs` and `design` are siblings that only depend on `proposal`, not a linear sequence — and how the twelve skills map onto it.

### Supporting changes

- `README.md`, `CONTEXT.md`, `CONTRIBUTING.md`, `scripts/validate_skill_catalog.py`, and `.claude-plugin/marketplace.json` all learn about the new `refinement` category (39 → 54 model-invoked skills).
- `skills/meta/ask-rhdh/SKILL.md` gets routing entries for all fourteen new skills.
- `skills/meta/setup-rhdh-skills/assets/catalog.json` gets the corresponding catalog entries and dependency edges (each artifact-creating skill depends on `rhdh-spec-driven-schema`; `apply-change` also needs `openspec-journal`; `archive-change` optionally uses `openspec-sync-specs`).

### Test plan

- `uv run pytest` — 377 passed.


skills/refinement/openspec-propose/SKILL.md
```diff
@@ -0,0 +1,63 @@
+---
+name: openspec-propose
+description: >-
+  Proposes a new OpenSpec change and generates every artifact it needs for
+  implementation in one step, framed as the friendly "describe what you want,
+  get a complete proposal back" entry point. Use for "propose a change for
+  X", "I want to build X, write it up", or "give me a proposal with design and
+  tasks ready to implement" — especially as someone's first change. Mechanics
+  are identical to openspec-ff-change; this skill exists for the softer
```
schultzp2020: This description says the mechanics are identical to `openspec-ff-change`. They are: same six steps, same output line. Two pointers for one write confuse the router and cost context on every turn. Keep one all-artifacts skill whose description covers both "propose a change" and "generate all artifacts". Delete the other directory.

skills/refinement/openspec-new-change/SKILL.md
```diff
@@ -0,0 +1,70 @@
+---
+name: openspec-new-change
+description: >-
+  Starts a new OpenSpec change using the rhdh-spec-driven artifact workflow
+  and shows the first artifact's template without writing it. Use for "start
+  a new change", "create a change for X", "I want to build/fix Y" when
```
schultzp2020: "I want to build/fix Y" is also how people reach propose. New-change writes nothing. Propose writes every artifact. That misroute is the painful kind. Claim only "scaffold a change" and "show the first template". Leave "I want to build X" to the all-artifacts skill.

skills/reference/openspec-journal/scripts/openspec-journal.py
```diff
@@ -0,0 +1,570 @@
...
+        "",
+        "Length handling:",
+        f"  input/output rejected at >{MAX_SUMMARY} chars (newlines stripped).",
+        "  Use --allow-truncate to force-write a truncated value.",
+    ]
+    return "Event vocabulary:\n" + "\n".join(rows) + "\n" + "\n".join(extras)
+
+
+def journal_path(root: Path, change: str) -> Path:
+    change_dir = root / "openspec" / "changes" / change
```
schultzp2020: The code joins `change` straight onto `openspec/changes/`. From a workspace with `openspec/changes/`, `../../secrets` appended `<workspace>/secrets/journal.jsonl` when that directory existed. Reject separators and `..` before you look anything up.

```suggestion
    if "/" in change or "\\" in change or ".." in change:
        raise SystemExit(
            f"Error: change name {change!r} must be a single path component."
        )
    change_dir = root / "openspec" / "changes" / change
```

Then require `change_dir.resolve()` to stay under `openspec/changes` (active or archive). Match date-prefixed archives with `YYYY-MM-DD-<name>`, not `endswith`.

skills/reference/openspec-journal/scripts/openspec-journal.py
```diff
@@ -0,0 +1,570 @@
...
+        fail(
+            2,
+            f"Error: {details} chars exceed limit ({MAX_SUMMARY}).\n"
+            "  Nothing written. Shorten the summary, or move long content\n"
+            "  to design.md or an ADR and reference it:\n"
+            '    output="See design.md §3 for full rationale."\n'
+            "  Force-write the truncated value with --allow-truncate.",
+        )
+    for key, val in kv.items():
+        record[key] = val
```
schultzp2020, line 442: I ran `decision input=x output=y event=archive ts=1970-01-01T00:00:00Z`. Exit 0. The stored line was `event: archive` with the forged timestamp. The same leftover loop accepted extra header-shaped fields, `mode=banana` on `task.complete`, and `count=nope` on `agent.spawned`. Freeze `ts`/`event`/`phase`. Allow only documented extras.

```suggestion
    allowed_extra = {"agents", "ctx_pct"}
    for key, val in kv.items():
        if key in {"ts", "event", "phase"}:
            fail(2, f"Error: cannot override '{key}' via k=v.")
        if key not in allowed_extra:
            fail(2, f"Error: unknown field '{key}'.")
        record[key] = val
```

Put `task.complete` in `MODE_VALUES` next to `mode.chosen`, and coerce `count` to int.

skills/reference/openspec-journal/scripts/openspec-journal-precompact-hook.sh
```diff
@@ -0,0 +1,64 @@
...
+
+set -u
+
+repo_root="$(git rev-parse --show-toplevel 2>/dev/null || true)"
+[ -n "${repo_root}" ] || exit 0
+
+helper="${repo_root}/scripts/openspec-journal.py"
+changes_dir="${repo_root}/openspec/changes"
+agent="${OPEN_SPEC_JOURNAL_AGENT:-Claude Code}"
+[ -x "${helper}" ] && [ -d "${changes_dir}" ] || exit 0
```
schultzp2020: This looks for `$git-root/scripts/openspec-journal.py`. The file is never shipped there, so the hook exits 0 and never logs `context.compacted`. Resolve the helper next to this script. Invoke it with `python3` (line 59 still runs `"${helper}"` as an executable).

The ranking `find` uses `stat -f`. On GNU coreutils that flag means `--file-system`, so the pipeline never yields a journal path. The mtime check below already has a `stat -c` fallback. The `find` does not.

```suggestion
helper="$(dirname "$0")/openspec-journal.py"
changes_dir="${repo_root}/openspec/changes"
agent="${OPEN_SPEC_JOURNAL_AGENT:-Claude Code}"
[ -f "${helper}" ] && [ -d "${changes_dir}" ] || exit 0
```

skills/refinement/openspec-new-change/SKILL.md
```diff
@@ -0,0 +1,70 @@
...
+then stop — drafting content is a separate turn.
+
+## Steps
+
+1. **Get the change name.** If the user's request already names it or
+   describes what they want to build, derive a kebab-case name (e.g. "add
+   user authentication" -> `add-user-auth`). Otherwise ask what they want to
+   build or fix; do not proceed without knowing.
+2. **Pick the schema.** Omit `--schema` to use the default (`rhdh-spec-driven`
+   in this repository) unless the user names a specific schema, or asks what
```
schultzp2020: Omitting `--schema` is said to use `rhdh-spec-driven` "in this repository". OpenSpec loads schemas from the project's `openspec/` (or the CLI package), not from this skill directory. Nothing copies `config.yaml` or `schemas/rhdh-spec-driven/` into the user's repo. First `openspec new change --schema rhdh-spec-driven` in a product checkout will not see Canonical Touchpoints or the journal obligation. Add an explicit install/copy step, or stop claiming that default until the schema is on disk.

skills/reference/rhdh-spec-driven-schema/schemas/rhdh-spec-driven/schema.yaml
```diff
@@ -0,0 +1,324 @@
+name: rhdh-spec-driven
+version: 1
+description: Default OpenSpec workflow - proposal → specs → design → tasks
```
schultzp2020: This is what `openspec schemas` shows. Specs and design each require only `proposal`. Your workflow doc already draws that sibling graph. A linear description makes continue-change's "first ready one" plus "never out of order" treat YAML list order as a dependency.

```suggestion
description: Default OpenSpec workflow - proposal → {specs, design} → tasks
```

Mirror that in the schema skill's always-loaded description. The body at line 29 already has the right form.

skills/refinement/openspec-ff-change/SKILL.md
```diff
@@ -0,0 +1,64 @@
...
+"All artifacts created! Ready for implementation.", and: "Run
+`/openspec-apply-change` to start working on the tasks."
+
+## Guardrails
+
+- Create every artifact `applyRequires` needs — not a subset.
+- Always read dependency artifacts before creating the next one.
+- Verify each artifact file exists after writing, before moving on.
+- If a change with that name already exists, suggest
+  `/openspec-continue-change` on the existing change instead of overwriting it.
```
schultzp2020: The description says this skill fast-forwards an existing change. Step 2 says skip `openspec new change` if the folder exists. This guardrail then sends that case to `/openspec-continue-change`, which writes one artifact. On an existing name, fill remaining `applyRequires`. Do not send the user to continue-change.

skills/refinement/openspec-continue-change/SKILL.md
```diff
@@ -0,0 +1,59 @@
...
+   ```
+
+   Read `schemaName`, `artifacts[].status`, and `isComplete`.
+3. **Act on status:**
+   - **All complete (`isComplete: true`)** — congratulate, show final status,
+     suggest implementing (`/openspec-apply-change`) or archiving
+     (`/openspec-archive-change`). Stop.
+   - **An artifact is `ready`** — pick the first ready one and create it by
+     following the artifact creation loop in `/rhdh-spec-driven-schema`
+     (`references/artifact-loop.md`). Create exactly one artifact, then stop.
```
schultzp2020: The caller should invoke `/rhdh-spec-driven-schema` and stop. `references/artifact-loop.md` lives in that skill, not here. After install that path is gone, so the loop never loads. Propose and ff-change copy the same punch-through. Drop the parenthetical and let the schema skill disclose the file.

skills/reference/rhdh-spec-driven-schema/config.yaml
```diff
@@ -0,0 +1,79 @@
...
+  Canonical long-lived feature and domain specs in `openspec/specs/`:
+  - capability specs that describe enduring system behavior
+  - the long-term destination for feature-level implementation canon
+
+  Change-local OpenSpec workspaces live in `openspec/changes/<change-name>/`.
+
+  ## Journal obligation (all phases)
+
+  Every change has an append-only `journal.jsonl`. Emit interaction events
+  via `scripts/openspec-journal.py` whenever something load-bearing happens
```
schultzp2020: This block is injected into product repos as OpenSpec context. `scripts/openspec-journal.py` will not be on that repo. The schema skill already says mechanics belong to `/openspec-journal`. Invoke that skill by name and delete the sibling script path. The apply Journal block in `schema.yaml` copies it too.

schultzp2020 (changes requested):
The catalog and CI look fine. A few first-run holes will bite people, so I would fix those before merge.

Routing is the expensive one. `openspec-propose` and `openspec-ff-change` are the same write with two always-loaded descriptions. `openspec-new-change` also claims "I want to build Y". That is how someone asks for a full proposal. If new-change wins, they get an empty folder. Keep one all-artifacts skill. Make new-change only about scaffolding.

The journal helper cannot run where the skills tell the agent to run it (`python3 scripts/openspec-journal.py` from the OpenSpec project). The file lives in the skill directory. Leftover `k=v` can overwrite `event` and `ts`. `../../secrets` writes outside `openspec/changes/`. `auth` attaches to an archived `user-auth`. The PreCompact hook looks under `$git-root/scripts/` and ranks journals with macOS `stat -f`, so it is a no-op here. Add tests for path containment, leftover fields, archive naming, and mode/count validation. Without those this CLI will drift the next time someone adds an event.

Nothing copies `rhdh-spec-driven` into a product repo's `openspec/`. `openspec new change --schema rhdh-spec-driven` will not see Canonical Touchpoints. The schema description is still `proposal -> specs -> design -> tasks` even though your workflow doc and the YAML `requires` graph treat specs and design as siblings.

Also: README says 54 model-invoked skills (catalog has 55) and "two" human-invoked entry points in the meta row. `AGENTS.md` still lists six folders. That list is what the next skill author will copy, so this PR should add `refinement/` there (and in ADR-0009). Onboard teaches a thinner cycle and archives with `openspec archive` while archive-change uses `mkdir`/`mv`. Pick one archive path. Bulk-archive inlines sync instead of invoking `/openspec-sync-specs`. Drop "reviewing an OpenSpec pull request" from audit so it does not fight `/rhdh-pr-review`.

gabemontero:
> Also: README says 54 model-invoked skills (catalog has 55) and "two" human-invoked entry points in the meta row. `AGENTS.md` still lists six folders. That list is what the next skill author will copy, so this PR should add `refinement/` there (and in ADR-0009). Onboard teaches a thinner cycle and archives with `openspec archive` while archive-change uses `mkdir`/`mv`. Pick one archive path. Bulk-archive inlines sync instead of invoking `/openspec-sync-specs`. Drop "reviewing an OpenSpec pull request" from audit so it does not fight `/rhdh-pr-review`.

@schultzp2020 can you highlight were in the `/rhdh-pr-review` skill it handles advesarial reviews for openspec content?

At least with my scans of rhdh-pr-review, I'm not seeing where it has the openspec specific directions we've curated 

I can tell you that using the openspec audit skill for reviewing openspec proposals in other's PRs has become a common task in the work currently occurring in the boost workspace 

@mareklibra FYI (in case you have cycles to compare rhdh-pr-review with what you curated for openspec-audit

thanks @schultzp2020

schultzp2020:
> > Also: README says 54 model-invoked skills (catalog has 55) and "two" human-invoked entry points in the meta row. `AGENTS.md` still lists six folders. That list is what the next skill author will copy, so this PR should add `refinement/` there (and in ADR-0009). Onboard teaches a thinner cycle and archives with `openspec archive` while archive-change uses `mkdir`/`mv`. Pick one archive path. Bulk-archive inlines sync instead of invoking `/openspec-sync-specs`. Drop "reviewing an OpenSpec pull request" from audit so it does not fight `/rhdh-pr-review`.
> 
> @schultzp2020 can you highlight were in the `/rhdh-pr-review` skill it handles advesarial reviews for openspec content?
> 
> At least with my scans of rhdh-pr-review, I'm not seeing where it has the openspec specific directions we've curated
> 
> I can tell you that using the openspec audit skill for reviewing openspec proposals in other's PRs has become a common task in the work currently occurring in the boost workspace
> 
> @mareklibra FYI (in case you have cycles to compare rhdh-pr-review with what you curated for openspec-audit
> 
> thanks @schultzp2020

@gabemontero I'll update the skill to include mentioning adversarial review and some other niceties I've been recently using alongside `/rhdh-pr-review`.

gabemontero:
> > > Also: README says 54 model-invoked skills (catalog has 55) and "two" human-invoked entry points in the meta row. `AGENTS.md` still lists six folders. That list is what the next skill author will copy, so this PR should add `refinement/` there (and in ADR-0009). Onboard teaches a thinner cycle and archives with `openspec archive` while archive-change uses `mkdir`/`mv`. Pick one archive path. Bulk-archive inlines sync instead of invoking `/openspec-sync-specs`. Drop "reviewing an OpenSpec pull request" from audit so it does not fight `/rhdh-pr-review`.
> 
> > 
> 
> > @schultzp2020 can you highlight were in the `/rhdh-pr-review` skill it handles advesarial reviews for openspec content?
> 
> > 
> 
> > At least with my scans of rhdh-pr-review, I'm not seeing where it has the openspec specific directions we've curated
> 
> > 
> 
> > I can tell you that using the openspec audit skill for reviewing openspec proposals in other's PRs has become a common task in the work currently occurring in the boost workspace
> 
> > 
> 
> > @mareklibra FYI (in case you have cycles to compare rhdh-pr-review with what you curated for openspec-audit
> 
> > 
> 
> > thanks @schultzp2020
> 
> 
> 
> @gabemontero I'll update the skill to include mentioning adversarial review and some other niceties I've been recently using alongside `/rhdh-pr-review`. 

Cool @schultzp2020 - if you remember tag us in that one - thanks

schultzp2020:
@gabemontero @mareklibra https://github.com/redhat-developer/rhdh-skills/pull/91

yangcao77:
@schultzp2020 
Good catches, thanks. worked through all of them.

The journal script is fixed as per commented. 
   
about the routing: merged `propose` into `ff-change`, Also fixed the schema description (`specs/design` are siblings), dropped the "default here" claim since nothing installs it, and cleaned up the docs 

Left the audit PR-review line as-is, per the thread with @gabemontero. waiting for @gabemontero and @mareklibra to confirm it's okay to remove that line.

gabemontero:
OK @yangcao77 @schultzp2020 @mareklibra I switched to opus 4.8 for this, then after a few go arounds with it to get primed on various aspects, zero'ed in on my original curiosity, and got an answer from opus 4.8 that I a) believe, and b) satisfies my curiosity and dismisses any concern that may have arisen if the answer to my curiosity's question landed in a particular direction.

Bottom line - I'm good with the removal of the text @schultzp2020 cited.

The tl;dr from opus 4.8 if you are curious:

```
  Paul wants the clause "or when reviewing an OpenSpec pull request" removed. Here's what that actually does.

  a) Precise impact of dropping the clause

  The description field is routing metadata, not behavior. It's the always-loaded text Claude reads to decide which skill to auto-invoke for a user utterance. The skill's Steps, gates, checklist, ownership map, fail-closed logic — none of that is touched. So the impact is surgical and confined to automatic model routing / discoverability:

  - Before: a request like "review this OpenSpec PR" matches two skills — openspec-audit-change (via this clause) and rhdh-pr-review (whose description already owns "review this PR", a GitHub PR URL/number). Two always-loaded descriptions competing for the same utterance. That's the "fight" — a routing collision where the wrong skill can win, e.g. someone wanting a code review of a PR gets the artifact auditor, or vice-versa.
  - After: the phrase "reviewing a PR" resolves cleanly to rhdh-pr-review; audit is triggered by "audit this change" / "run the audit". This is the same split-by-verb discipline Paul applied in PR #91 (review-the-code vs. test-on-cluster): one verb, one skill.
  - Why the clause was there at all: it's inherited verbatim from the a/b standalone origin ("...or when reviewing an OpenSpec PR"), where there was no rhdh-pr-review to collide with. It only becomes a problem once the audit skill is merged into the same repo as Paul's PR-review skill. That's the whole crux of his comment.

  There's also a subtle correctness angle in his favor: the audit skill doesn't review a PR at all — it has no forge-fetch step. It audits a change directory on disk (openspec/changes/<name>/) via openspec list/status or filesystem fallback. So "when reviewing an OpenSpec pull request" was always a slightly false advertisement of what the skill does, which is a second reason to drop it.

  b) If you explicitly apply the audit skill to another user's PR, does removing this line prevent you?

  No. You are not prevented in any way.

  - The description governs only automatic selection. Explicit invocation bypasses it entirely — typing /openspec-audit-change (or /opsx:audit), or "use the audit skill on this," calls it by name. The repo's own README states model-invoked skills "can also be called by name." Removing a routing-hint clause cannot disable a named invocation.
  - None of the skill's operational contract changes: same Steps, same apply-ready gate, same A–H checklist, same audit.md output.

  The one thing to be clear about — and it's a practical prerequisite unrelated to the removed line — is what you're pointing it at. The audit skill reads a change from the local filesystem / openspec CLI, not from GitHub. So "apply audit to another user's PR" really means: check out that PR's branch (so openspec/changes/<name>/ is on disk), then run /openspec-audit-change. That requirement exists identically whether or not the clause is present. Paul's edit
  removes an auto-trigger phrase; it removes zero capability.

  Bottom line: dropping the line only stops the audit skill from auto-claiming "review this PR" requests. Your ability to deliberately run the audit against a checked-out PR's change artifacts is fully intact — the real gate is having those artifacts locally, not the description text.

```

yangcao77:
@gabemontero thanks for revisit & confirming that. I've dropped the word `or when reviewing an OpenSpec pull request` to address Paul's concern on having conflict with the pr-review skill. I believe this should not affect the purpose and usage of the audit skill to keep context consistency in the feature refinement workflow

skills/refinement/openspec-new-change/SKILL.md
```diff
@@ -0,0 +1,75 @@
...
+   describes what they want to build, derive a kebab-case name (e.g. "add
+   user authentication" -> `add-user-auth`). Otherwise ask what they want to
+   build or fix; do not proceed without knowing.
+2. **Pick the schema.** Run `openspec schemas --json` to see what this
+   project actually offers. `rhdh-spec-driven` is the default only when the
+   repo's `openspec/` has it configured — its `config.yaml` and
+   `schemas/rhdh-spec-driven/` files live in `/rhdh-spec-driven-schema` and
+   must be present under the repo's `openspec/` for `--schema rhdh-spec-driven`
+   to resolve. If it is not listed, name an available schema instead. Omit
+   `--schema` to take the configured default.
```
schultzp2020: This caveat is the gap. OpenSpec resolves schemas from the product repo `openspec/schemas/`, not from this skill directory. `/setup-rhdh-skills` does not copy `config.yaml` or `schemas/rhdh-spec-driven/` into a checkout. `openspec-ff-change` and `openspec-onboard` then run `openspec new change` without `--schema`.

A first `openspec new change --schema rhdh-spec-driven` in rhdh will not see Canonical Touchpoints or the sibling specs/design graph. Add a copy or `openspec schema` step from these files, or stop calling `rhdh-spec-driven` the default until it is on disk.

skills/reference/openspec-journal/scripts/openspec-journal.py
```diff
@@ -0,0 +1,603 @@
...
+
+def journal_path(root: Path, change: str) -> Path:
+    if "/" in change or "\\" in change or ".." in change:
+        raise SystemExit(
+            f"Error: change name {change!r} must be a single path component "
+            "(no '/', '\\', or '..')."
+        )
+    change_dir = root / "openspec" / "changes" / change
+    if change_dir.is_dir():
+        return change_dir / "journal.jsonl"
```
schultzp2020: The `/`, `\`, and `..` check is the right idea, but `is_dir()` still accepts `.` and `archive`. I ran `decision` with change `.` and with `archive`. Both exited 0 and wrote `openspec/changes/journal.jsonl` and `openspec/changes/archive/journal.jsonl`. A symlink under `openspec/changes/<name>` follows out of tree the same way.

After you resolve the change directory, require it stay under `openspec/changes` and not be `openspec/changes` or `openspec/changes/archive` itself.

`--input-file` and `--output-file` still read any path into the JSONL line. I pointed `--input-file` at a file outside the workspace and the token landed in `input`. Bound those paths to the change directory. Freeze `phase` with `ts` and `event`.

`SKILL.md` still tells the agent to run `python3 scripts/openspec-journal.py` from the product repo. The hook now resolves the helper next to itself. This invocation should too.

skills/refinement/openspec-ff-change/SKILL.md
```diff
@@ -0,0 +1,68 @@
...
+description: >-
+  Proposes an OpenSpec change and generates every artifact it needs for
+  implementation — proposal, specs, design, tasks — in one pass, for a
+  brand-new or an existing change, with progress tracked via TodoWrite. Use
+  for "propose a change for X", "I want to build X, write it up", "give me a
+  proposal with design and tasks ready to implement", "fast-forward this
+  change", "generate all the artifacts now", or "get this ready to
+  implement" — including someone's first change. For one artifact at a time,
+  see openspec-continue-change; to scaffold the folder without writing
+  anything, see openspec-new-change.
```
schultzp2020: This description still owns "propose a change for X" and "including someone's first change". Those are how people reach onboard, and how they ask for a full proposal. If ff-change wins on a first-time user, they get every artifact dumped instead of a walkthrough.

Claim only fast-forward / generate every remaining artifact. Drop the sibling names `openspec-continue-change` and `openspec-new-change`. CONTRIBUTING.md says the description is the routing surface, and a sibling token fires on presence.

skills/refinement/openspec-continue-change/SKILL.md
```diff
@@ -0,0 +1,59 @@
+---
+name: openspec-continue-change
+description: >-
+  Continues an OpenSpec change by creating exactly the next ready artifact,
+  one at a time. Use for "continue this change", "what's next", "create the
+  next artifact", or when the user wants to progress a change step by step
+  rather than generating everything at once. Stops after one artifact; use
+  openspec-ff-change instead when the user wants every remaining artifact
+  created in one pass.
```
schultzp2020: "continue this change" is also how people mean keep implementing. Apply-change claims "start/continue implementing" and "keep going on <change>". One writes the next artifact. The other writes code. That misroute is the expensive kind.

Claim only the next ready artifact, one at a time. Leave "continue implementing" to apply-change. Same CONTRIBUTING.md rule: do not name `openspec-ff-change` here.

skills/refinement/openspec-onboard/SKILL.md
```diff
@@ -0,0 +1,163 @@
...
+|---|---|
+| `/openspec-ff-change` | Propose a change and generate all artifacts in one pass |
+| `/openspec-explore` | Think through problems before/during work |
+| `/openspec-apply-change` | Implement tasks from a change |
+| `/openspec-archive-change` | Archive a completed change |
+| `/openspec-new-change` | Start a new change, step through artifacts one at a time |
+| `/openspec-continue-change` | Continue working on an existing change |
+| `/openspec-audit-change` | Audit artifacts for cross-artifact consistency (pre-implementation) |
+| `/openspec-verify-change` | Verify implementation matches artifacts (post-implementation) |
+| `/grilling` (external, optional) | Resolve a genuinely open design question before moving to tasks |
```
schultzp2020, line 134: This recap is what a first-time user will copy. `/openspec-new-change` scaffolds the folder and stops. It does not step through artifacts one at a time. That is continue-change.

Phases 5 through 9 also inline proposal, specs, design, tasks, and apply instead of invoking those skills. Onboard then teaches a linear `proposal -> specs -> design -> tasks` cycle while the schema graph treats specs and design as siblings.

Have each phase invoke the real skill by name, and make this table match their actual triggers.

schultzp2020 (changes requested):
Thanks for the last round. Catalog membership and CI look fine. The inlines are the remaining first-run holes.

Also, `docs/refinement/refinement-workflow.png` is about 2.8MB and unused. The markdown already embeds the jpeg. ADR-0009 consequences still list five installer groups after this PR adds `refinement`.

skills/refinement/openspec-new-change/SKILL.md
yangcao77: Added `install_project_schema.py` to copy the config and schema into the product repo’s `openspec/`.
`/rhdh-spec-driven-schema`, `/openspec-new-change`, `/openspec-ff-change`, `/openspec-onboard`, and `setup install.md`all call this at startup/install.

skills/reference/openspec-journal/scripts/openspec-journal.py
yangcao77: new commit now fixes this. 
it rejects `. ` & `archive`, confines resolved change dirs under `openspec/changes/.` bounds `--input-file` & `--output-file` to the change dir. freezes phase with `ts` and `event`.
`SKILL.md` now invokes the helper from the skill install path, not the product repo.

skills/refinement/openspec-ff-change/SKILL.md
yangcao77: ff-change: only fast-forward / generate remaining artifacts.
continue-change: only next ready artifact; no “continue implementing”; no sibling skill names.

skills/refinement/openspec-continue-change/SKILL.md
yangcao77: ff-change: only fast-forward / generate remaining artifacts.
continue-change: only next ready artifact; no “continue implementing”; no sibling skill names.

skills/refinement/openspec-onboard/SKILL.md
yangcao77: new-change description is now scaffold-only
the onboard  invokes real skills by name; also now noted in phases that  `specs` and `design` are siblings, and can be invoke in either order (or parallel) as they only depend on proposal, as documented in the workflow doc & in the config/schema

schultzp2020 (approved):
/lgtm

yangcao77:
@schultzp2020 do you mind helping merge this PR? I do not have permission to do so

PR #94, opened 2026-08-26
feat(skill-authoring): add harness-native evaluations

schultzp2020:
## Hey, I just made a Pull Request!

Applies the evaluation guidance from [Don't Ship Skills Without Evals](https://www.youtube.com/watch?v=0vphxNt4wyk), [OpenAI's skill evaluation workflow](https://developers.openai.com/blog/eval-skills), and [Anthropic's test design guidance](https://platform.claude.com/docs/en/test-and-evaluate/develop-tests), using [agent-eval-harness](https://github.com/opendatahub-io/agent-eval-harness) as the executable standard.

This makes skill evaluation part of creation, audit, consolidation, and retirement instead of treating a well-formed `SKILL.md` as evidence that the skill improves agent behavior.

This diagram shows the two contracts independently: prompt mode measures whether the agent selects the skill, while skill mode measures what happens after explicit invocation.

```mermaid
flowchart LR
  Prompts["Explicit, implicit, and near-miss prompts"] --> Routing["Prompt-mode routing eval"]
  Routing --> Selection["Skill selected only when expected"]
  Fixtures["Create, audit, and consolidate fixtures"] --> Behavior["Skill-mode behavior eval"]
  Behavior --> Checks["Deterministic checks"]
  Behavior --> Judge["Read-only semantic judge"]
  Selection --> Baseline["Comparable baseline and regression thresholds"]
  Checks --> Baseline
  Judge --> Baseline
```

### What changed

- `skill-authoring` now defines evaluation across routing, outcome, process, style, privacy, and efficiency. It distinguishes temporary capability skills from durable preference skills and requires evidence before shipping or retiring behavior.
- Agent Eval Harness owns the mechanics. The authoring workflow uses its `eval.yaml`, isolated case directories, deterministic and semantic judges, repeated samples, baselines, and thresholds instead of introducing a second skill-local manifest.
- The routing suite contains twelve prompt-mode cases spanning explicit, implicit, contextual, negative, and keyword-sharing requests.
- The behavior suite contains six skill-mode cases covering create, audit, and consolidate, including a consolidation that should be rejected because the skills share a noun but not a verb.
- Codex is the committed primary runner. Claude Code comparisons reuse the same datasets rather than maintaining runner-specific copies.

### Compatibility and execution

The runtime skill remains portable under the Agent Skills standard. Agent Eval Harness is the default evaluation integration when available; a repository that mandates another evaluator can keep that host contract. If no runner is available, authors may prepare harness-native assets but must report the exact blocker and leave the behavior marked unevaluated.

The two committed suites are intentionally marked unevaluated. Their configs were validated against the upstream harness, but no authenticated run used a project-staged copy of this changed skill and its external `/grilling` dependency. The semantic threshold is provisional until the first successful baseline calibrates it; repository tests and prose review are not presented as behavioral evidence.


04kash:
Thanks for setting this up! In terms of the goals I'd want the evals to meet, this is what I was thinking:

1. Safety Evals
- Untrusted Skill, Trusted User: Checks for prompt injection, exfiltration, malware, or supply-chain risks. (Some of this may already be covered by skills.sh publishing audits)
- Trusted Skill, Untrusted User: Resiliency against adversarial exploitation. Prioritized order: Accidental misuse > Indirect injection via external inputs > Direct adversarial prompting.

2. Behavioral Evals (Action Items for this PR)
- Baseline Arm: Currently, the behavior suite runs with-skill only. We need to add a without-skill baseline arm across the same test cases to measure true Skill Lift.
- Uplift Measurement: Dependent on the baseline arm above. Once both arms exist, /eval-compare should give us per-case Skill Lift.
- Context Efficiency: We should track context impact starting with Activation Rate, followed by Token Budget, and finally Signal-to-Noise Ratio.

I think we should address the Baseline Arm and Uplift Measurement in this PR so we can establish Skill Lift baselines right away. We could track the remaining safety and context-efficiency items in a follow-up PR.

PR #99, opened 2026-09-01
feat(release): add rhdh-smoke-tests Helm and Operator smoke-test skill

nickboldt:
## Summary
- Add the `/rhdh-smoke-tests` skill under `skills/release/` for Helm and Operator RC smoke tests.
- One namespace per path: install previous GA, upgrade to this RC, then (older stream only) upgrade to the latest next-minor GA.
- Guest via Helm `upstream.backstage.appConfig` or an Operator ConfigMap; `verify_ns.sh` and `operator_upgrade.sh` for the deterministic steps.

## Test plan
- [ ] `uv run pytest tests/unit/test_operator_upgrade.py tests/unit/test_verify_ns.py tests/unit/test_skill_catalog.py tests/unit/test_ask_rhdh_routes.py`
- [ ] `python scripts/validate_skill_catalog.py` reports valid
- [ ] `bash -n` on `skills/release/rhdh-smoke-tests/scripts/*.sh`

Ref: https://redhat.atlassian.net/browse/RHIDP-16678

Generated-by: cursor


skills/release/rhdh-smoke-tests/scripts/operator_upgrade.sh
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env bash
```
schultzp2020: Can we switch the scripts to python?

nickboldt: yes we COULD switch to python but these scripts wrap and extend bash scripts in https://github.com/redhat-developer/rhdh-operator/blob/main/.rhdh/docs/installing-ci-builds.adoc so I'd argue we should be consistent with bash+jq instead of reimplementing in python and then having to support 2 languages

schultzp2020 (approved):
/lgtm

PR #103, opened 2026-09-02
fix: point generated Jira URLs to redhat.atlassian.net

psrna:
JIRA_BASE and JIRA_SEARCH_BASE were hardcoded to issues.redhat.com (Jira Data Center). All browse and search links now resolve to the Jira Cloud instance. Fixed Dashboard URLs/IDs


invincibleJai (approved):
/lgtm

PR #105, opened 2026-09-02
chore: update rhdh-backstage-upgrade skill to effectively bump rhdh monorepo

hopehadfield:
## Summary                                                                                                                                                                    
                                                                                                                                                                                 
   Expand `/rhdh-backstage-upgrade` to support Backstage version bumps in the `redhat-developer/rhdh` monorepo while preserving its existing plugin and standalone app           
 workflows.                                                                                                                                                                      
                                                                                                                                                                                 
   This addresses the upgrade gaps identified in #106 without adding a separate monorepo-specific skill.                                                                         
                                                                                                                                                                                 
   ## Changes                                                                                                                                                                    
                                                                                                                                                                                 
   - Use a repository-provided `versions:bump` wrapper when available.                                                                                                           
   - Validate target versions before passing them to the Backstage CLI.                                                                                                          
   - Preserve the repository's existing dependency range style after a bump.                                                                                                     
   - Audit `resolutions` and `overrides` against their original issue, advisory, or patch rationale.                                                                             
   - Stop when an install, migration, or required dedupe step fails instead of continuing from a partial dependency state.                                                       
   - Add guidance for new or renamed backend service factories.                                                                                                                  
   - Select compatible `@red-hat-developer-hub/*` package versions from peer ranges and release metadata.                                                                        
   - Add the full `redhat-developer/rhdh` verification sequence, including lint, formatting, monorepo checks, and Dockerfile generation.                                         
   - Keep the generic plugin workflow available for individual plugins, workspaces, and standalone Backstage apps.                                                               
   - Update the generated `/ask-rhdh` route description.


skills/release/rhdh-backstage-version-bump/SKILL.md
```diff
@@ -0,0 +1,233 @@
...
+
+Confirm the local checkout is `redhat-developer/rhdh` on `main`. Read
+`backstage.json` and the root `package.json`. If the repository is already on
+the target version, report that and stop before making changes.
+
+### 2. Run the bump script
+
+```bash
+yarn versions:bump                  # latest
+yarn versions:bump --release 1.56.0 # specific version
```
schultzp2020: The release value comes from free-form issue text. However, `scripts/update-backstage.mjs` interpolates it into a command string passed to `execSync`. Shell syntax in that value would execute in the checkout. Validate it as an exact stable semver, then confirm that the release exists. Never forward the raw issue value to this command.

skills/release/rhdh-backstage-version-bump/SKILL.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Procedure
+
+### 1. Read the issue and confirm the target
+
+Use `/rhdh-forge` to fetch the GitHub issue that triggered Fullsend. Extract the
+requested Backstage version, linked Jira key if present, expected branch, and any
+special reviewer notes.
+
+Confirm the local checkout is `redhat-developer/rhdh` on `main`. Read
```
schultzp2020: This confirms `main`, but the workflow never creates a feature branch or checks the starting worktree. Following it literally leaves the commit on `main`. It cannot produce a `main`-targeted PR from a distinct head and can sweep existing edits into the bump. Before changing files, require the canonical remote, a clean worktree, a fresh `origin/main`, and an explicit head branch.

skills/release/rhdh-backstage-version-bump/SKILL.md
```diff
@@ -0,0 +1,233 @@
...
+yarn versions:bump --release 1.56.0 # specific version
+```
+
+This runs `backstage-cli versions:bump --skip-install`, pins all dependency
+ranges (removes `^` prefixes), runs `yarn install --no-immutable && yarn dedupe`,
+and updates `backstage.json` and `packages/app/src/build-metadata.json`.
+
+The script may fail partway through. If `backstage-cli versions:bump` itself
+fails, check whether the requested version exists. If `yarn install` fails after
+the bump, proceed to step 3 — the build loop will surface and fix the issues.
```
schultzp2020: The target repository's bump script updates `backstage.json` and build metadata only after `yarn install` and `yarn dedupe` succeed. Continuing after an install failure sends the build loop into a partial state. Its errors are not reliable migration evidence. Require a successful bump, install, and dedupe. Diagnose the install failure and rerun the bump before editing source.

skills/release/rhdh-backstage-version-bump/SKILL.md
```diff
@@ -0,0 +1,233 @@
...
+the target version, report that and stop before making changes.
+
+### 2. Run the bump script
+
+```bash
+yarn versions:bump                  # latest
+yarn versions:bump --release 1.56.0 # specific version
+```
+
+This runs `backstage-cli versions:bump --skip-install`, pins all dependency
```
schultzp2020: This loop duplicates `/rhdh-backstage-upgrade`, whose contract already owns Backstage version changes. The two paths have already diverged. The existing owner runs `versions:migrate` and reads every intervening release changelog, while this workflow does neither. Keep this skill's issue-to-PR orchestration. Invoke the upgrade owner and declare it as a catalog dependency so the workflow includes runtime and configuration migrations.

skills/release/rhdh-backstage-version-bump/SKILL.md
```diff
@@ -0,0 +1,233 @@
...
+export that is part of the default backend setup needs to be added to the
+`DEFAULT_SERVICE_FACTORIES` array. Keep the array alphabetically sorted and
+update the doc comment at the top of the file if the omission list changes.
+
+### Non-Backstage dependency breakage
+
+Build errors do not always trace to `@backstage/*` packages. Bundled
+`@red-hat-developer-hub/*` packages may also need bumping to their latest version
+to stay compatible with the new Backstage release. If a `ConfigSchemaError` or
+type error points to a `@red-hat-developer-hub/*` package, bump it to the latest
```
schultzp2020: The latest published RHDH package can target a newer Backstage line or add unrelated breaking changes. A type error does not establish that "latest" is compatible with this bump. Check peer ranges and release metadata, then choose the smallest package version compatible with the target Backstage release.

skills/release/rhdh-backstage-version-bump/SKILL.md
```diff
@@ -0,0 +1,233 @@
...
+packages to specific versions or patch locators. After a bump, review every
+entry and remove anything that is no longer relevant. A resolution is irrelevant
+when:
+
+- The pinned package is no longer in the dependency tree
+- The reason for the pin (a bug, a breaking change) has been resolved in the
+  bumped version
+- The resolution conflicts with what the bump introduced
+
+When in doubt, remove the resolution and see if `yarn install` and `yarn build`
```
schultzp2020: Install and build success do not prove that a resolution is obsolete. A pin can encode a runtime, security, or regression workaround that these checks never exercise. Trace each resolution to its history or issue. Remove it only when that reason no longer applies, then run the relevant regression check. Do not use "when in doubt, remove" as the default.

schultzp2020 (changes requested):
I found several issues to resolve before this workflow is safe and complete.

skills/plugins/rhdh-backstage-upgrade/references/bump-deps.md
```diff
@@ -2,11 +2,21 @@
 
 ## Primary command
 
+Use a repository-provided wrapper when one exists; it may enforce pinning and
+update generated metadata that the CLI does not know about. In
+`redhat-developer/rhdh`, run:
```
schultzp2020, line 10: The linked issue requires discovering and bumping every independent workspace root, but this runs the RHDH wrapper only once from the repository root. At reproduced commit `83a4ee9d`, the root and `dynamic-plugins/` have separate `backstage.json` and `yarn.lock` files. Issue #106 documents that the root command left `dynamic-plugins/` on the old release. Discover every non-vendored `backstage.json` root, run the selected bump and migration phases for each, and verify every manifest and lockfile before completion.

skills/plugins/rhdh-backstage-upgrade/references/bump-deps.md
```diff
@@ -30,6 +40,22 @@ In a monorepo (like `rhdh-plugins`), run from the workspace root. The command up
...
+
+Record whether each dependency section uses exact versions, carets, tildes, or
+workspace ranges before the bump. `versions:bump` may rewrite that style. Restore
+the existing convention after the command, then review every changed dependency
+and the lockfile. In `redhat-developer/rhdh`, `@backstage/*` dependencies remain
+exact pins.
+
+## Audit resolutions and overrides
+
+Review changed packages against root `resolutions` and `overrides`. Trace each
```
schultzp2020: This limits the audit to resolutions related to changed packages. The reported stale `@scalprum/*` pins were transitive entries made obsolete by the new dependency graph, so they need not match a direct package changed by `versions:bump`. Enumerate every resolution and override in each discovered root, then trace its provenance. Compare the dependency graph before and after the bump. Record the reason for keeping or removing each entry.

schultzp2020 (changes requested):
The linked issue's independent-workspace path is still incomplete.

schultzp2020 (approved):
/lgtm

PR #111, opened 2026-09-08
feat(skills): add rhdh-catalog-index

gustavolira:
RHIDP-16848. Grew out of RHDHBUGS-3720.

## Why

Overriding the plugin catalog index came up twice in one week — 1.10.4 shipping a stale index, and mirroring for a disconnected install — and both times the procedure was worked out from scratch. Each of the three steps has a trap that costs about an hour.

## What the skill covers

**Resolving the digest.** The chart and the Operator both pin the *manifest list*, but the per-architecture child sits in the same tag listing, and pinning it locks the reference to one arch. `resolve_index_digest.py` prints both, labelled, and exits 2 when a deployment is behind:

```
manifest list       sha256:65a60ffc...
bare digest         65a60ffc...   <- Helm --set value
  child linux/amd64 sha256:ef20ffb6...   <- do NOT pin this
drift               YES - deployment is behind
```

It computes the list digest from the raw manifest rather than `skopeo inspect --format {{.Digest}}`, which resolves an image instance and fails outright on a darwin/arm64 workstation against a linux-only index.

**Writing the override.** Helm takes a bare digest, because `repository` already ends in `@sha256`. The Operator takes `extraEnvs.envs` — an object holding a list, not a list; a bare list is rejected by the CRD — scoped to `install-dynamic-plugins`, with no `oci://` prefix, on `v1alpha5`.

**Proving it took.** This is the part that needed a script. `install-dynamic-plugins.py` falls back from `registry.access.redhat.com/rhdh/` to `quay.io/rhdh/`, so a stale index installs cleanly, the pod goes green, and the user quietly gets older plugin builds. Plugins installing is not evidence. `verify_index.py` reports index used, plugins installed, and fallback count, and exits 2 on a mismatch.

## Verification

Both scripts were run against the live registry and against a captured `install-dynamic-plugins` log. The Operator half was verified end to end on OCP 4.22, two Backstage CRs in one namespace differing only by the override:

| | Index used | New digests | Old digests | quay.io fallbacks |
|---|---|---|---|---|
| No override | `51d12fc0` | 0 | 5 | 6 |
| With override | `65a60ffc` | 4 | 0 | 0 |

The Helm half was verified the same way the day before.

## Scope

Does not run e2e (`/e2e-deploy-rhdh`, `/rhdh-prow-trigger`), does not smoke test a release (`/rhdh-smoke-tests`), and does not fix an index at the source. Consumer-side override only.

`RELATED_IMAGE_catalog_index` through `Subscription.spec.config.env` would be a cluster-wide alternative to the per-CR override. The operator's default config names it, but nothing sets it and I have not verified it — the skill says that rather than recommending it as tested.


gustavolira:
Closing — the conclusion from the thread with Nick is that packaging the workaround is the release note in another format, and the check belongs earlier.

What survives is the assertion, not this skill: before installing, every ref the catalog index names must resolve from registry.access.redhat.com **without** falling back to quay. The fallback is why nobody noticed 1.10.4 — a stale index installs green either way, so a post-install check only tells you after the fact, on one cluster.

Moving that into `/rhdh-smoke-tests` as a pre-install gate. No cluster needed for it: it is a registry check on the index the release is about to ship. Not adding it to `/e2e-deploy-rhdh` — no need to re-check r.a.r.c on every CI build.

The pipeline-side equivalent is flagged on RHIDP-16624.

PR #114, opened 2026-09-08
fix: validate Prow triggers and add read-only status checks

nickboldt:
The nightly trigger CLI tells users to repeat a submission command to refresh status, which can create duplicate ProwJobs. It also accepts unrelated periodic job names and suggests an invalid job name after every API failure. Its dry-run output lacks the command, impact, and recovery details required by the skill workflow.

This PR makes nightly submissions easier to review and gives users a status command that cannot submit another job.

- Add `--status <EXECUTION_ID>` and print it after submission. Reject combinations with submission flags.
- Restrict job names to the supported nightly patterns. Before live submission, require membership in the owning repository's configured-job list. If job discovery cannot verify membership, stop submission.
- Give separate guidance for authentication, permissions, invalid requests, missing executions, rate limits, service failures, and network failures. Treat uncertain submission outcomes explicitly and avoid automatic POST retries.
- Expand offline previews with the executable command, overrides, validation limits, resource impact, unknown cost, abort guidance, and verification steps. Previewing neither accesses credentials nor creates configuration.

Update the skill workflow with status and recovery instructions, and add regression tests for these behaviors. The release-1.10 dry run with tag `1.10-171` and catalog image `registry.access.redhat.com/rhdh/plugin-catalog-index:1.10.4-1788447603` retains its original payload.

## Validation

- Full suite: 673 passed with `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_NOSYSTEM=1 UV_CACHE_DIR=/tmp/rhdh-prow-trigger-uv-cache uv run pytest`.
- Focused trigger tests: 44 passed.
- `uv run ruff check .` and `uv run ruff format --check .`: passed.
- Skill catalog validation and `git diff --check`: passed.
- Ran the exact image-verification command with `--dry-run`. The command submitted no ProwJob.

The test environment excludes global Git configuration so temporary fixtures do not inherit GPG signing. Earlier runs exceeded the unchanged prose-linter timing threshold. The final full run passes.

generated-by: codex using gpt-6-astra xhigh

### DIFF
--- a/skills/ci/rhdh-prow-trigger/scripts/gangway_adapter.py
+++ b/skills/ci/rhdh-prow-trigger/scripts/gangway_adapter.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 import json
 import subprocess
 import urllib.error
+import urllib.parse
 import urllib.request
 from typing import Any
 
@@ -19,6 +20,10 @@ GANGWAY_URL = "https://gangway-ci.apps.ci.l2s4.p1.openshiftapps.com/v1/execution
 
 class GangwayAdapterError(RuntimeError):
     """A credential-opaque failure from the Gangway adapter."""
+
+    def __init__(self, message: str, *, outcome_unknown: bool = False) -> None:
+        super().__init__(message)
+        self.outcome_unknown = outcome_unknown
 
 
 class GangwayAdapter:
@@ -29,18 +34,26 @@ class GangwayAdapter:
         self.executable = executable
 
     def _token(self) -> str:
-        result = subprocess.run(
-            [self.executable, "--kubeconfig", self.kubeconfig, "whoami", "-t"],
-            capture_output=True,
-            text=True,
-            encoding="utf-8",
-            errors="replace",
-            check=False,
-            shell=False,
-        )
+        try:
+            result = subprocess.run(
+                [self.executable, "--kubeconfig", self.kubeconfig, "whoami", "-t"],
+                capture_output=True,
+                text=True,
+                encoding="utf-8",
+                errors="replace",
+                check=False,
+                shell=False,
+            )
+        except OSError as error:
+            raise GangwayAdapterError(
+                "Cannot run oc. Run /setup-rhdh-skills openshift-ci, then retry."
+            ) from error
         token = result.stdout.strip()
         if result.returncode != 0 or not token:
-            raise GangwayAdapterError("OpenShift CI authentication is missing or expired")
+            raise GangwayAdapterError(
+                "OpenShift CI authentication is missing or expired. "
+                "Run /setup-rhdh-skills openshift-ci, then retry."
+            )
         return token
 
     def _request(
@@ -59,14 +72,52 @@ class GangwayAdapter:
         )
         try:
             with urllib.request.urlopen(request, timeout=30) as response:
-                return json.loads(response.read().decode("utf-8"))
+                body = json.loads(response.read().decode("utf-8"))
         except urllib.error.HTTPError as error:
-            raise GangwayAdapterError(f"Gangway returned HTTP {error.code}") from error
-        except (urllib.error.URLError, OSError, json.JSONDecodeError) as error:
-            raise GangwayAdapterError(f"Gangway request failed: {error}") from error
+            if error.code == 401:
+                guidance = (
+                    "Authentication was rejected. Run /setup-rhdh-skills openshift-ci, then retry."
+                )
+            elif error.code == 403:
+                guidance = "Permission denied. Ask an OpenShift CI administrator to check access."
+            elif error.code == 400:
+                guidance = "Invalid request. Check the job name, execution ID, and overrides."
+            elif error.code == 404:
+                guidance = (
+                    "Execution not found. Check the execution ID."
+                    if method == "GET"
+                    else "Job or endpoint not found. Check the configured job list and Gangway URL."
+                )
+            elif error.code == 429:
+                guidance = "Rate limit exceeded. Wait before retrying."
+            elif error.code >= 500:
+                guidance = "Gangway service failure. Check OpenShift CI service availability."
+            else:
+                guidance = "Check the request and OpenShift CI service availability."
+            raise GangwayAdapterError(
+                f"Gangway returned HTTP {error.code}. {guidance}",
+                outcome_unknown=method == "POST" and (error.code == 408 or error.code >= 500),
+            ) from error
+        except (urllib.error.URLError, OSError) as error:
+            raise GangwayAdapterError(
+                "Gangway network request failed. Check DNS, network/VPN connectivity, "
+                "and OpenShift CI service availability.",
+                outcome_unknown=method == "POST",
+            ) from error
+        except (json.JSONDecodeError, UnicodeDecodeError) as error:
+            raise GangwayAdapterError(
+                "Gangway returned an invalid JSON response. Check OpenShift CI service availability.",
+                outcome_unknown=method == "POST",
+            ) from error
+        if not isinstance(body, dict):
+            raise GangwayAdapterError(
+                "Gangway returned an unexpected response; expected a JSON object.",
+                outcome_unknown=method == "POST",
+            )
+        return body
 
     def trigger(self, payload: dict[str, Any]) -> dict[str, Any]:
         return self._request(GANGWAY_URL, method="POST", payload=payload)
 
     def status(self, job_id: str) -> dict[str, Any]:
-        return self._request(f"{GANGWAY_URL}/{job_id}")
+        return self._request(f"{GANGWAY_URL}/{urllib.parse.quote(job_id, safe='')}")
--- a/skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
+++ b/skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
@@ -46,6 +46,7 @@ import argparse
 import json
 import os
 import re
+import shlex
 import shutil
 import subprocess
 import sys
@@ -64,6 +65,22 @@ REPOS = [
 ]
 
 OVERLAY_JOB_PREFIX = "periodic-ci-redhat-developer-rhdh-plugin-export-overlays-"
+NIGHTLY_JOB_PATTERN = re.compile(
+    r"periodic-ci-redhat-developer-rhdh-(?:plugin-export-overlays-)?"
+    r"[a-zA-Z0-9][a-zA-Z0-9_.-]*-nightly"
+)
+TRIGGER_OVERRIDES = (
+    "image_registry",
+    "image_repo",
+    "tag",
+    "catalog_index_image",
+    "chart_version",
+    "playwright_version",
+    "org",
+    "repo",
+    "branch",
+    "send_alerts",
+)
 
 
 # --- Logging ---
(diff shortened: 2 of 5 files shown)


nickboldt:
For context I asked codex to review how we run e2e tests for pre-GA and post-GA smoke testing, and it came up with some concerns. If any of this seems valid, cool. If not we can close this as needless noise!

skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
```diff
@@ -403,49 +434,73 @@ def trigger_job(adapter: GangwayAdapter, payload: dict) -> dict:
...
             job_url = data.get("job_url", "")
             if job_url:
                 break
-        except GangwayAdapterError:
-            pass
-        time.sleep(2)
+        except GangwayAdapterError as error:
+            print("", file=sys.stderr)
+            log_warn(f"Job was submitted, but status lookup failed: {error}")
+            return
```
gustavolira, line 488: Returning on the first status error undoes what the retry loop is for. I poked the real Gangway with my CI session and it answers 500 even for IDs it does not know, so the first GET right after submission can easily fail once, and this kills polling on attempt 1 with a scary "service failure" message for a job that is actually fine. On main this survived transient errors for 5 attempts. Maybe keep retrying on 5xx/network and only bail on 401/403?

skills/ci/rhdh-prow-trigger/scripts/gangway_adapter.py
```diff
@@ -59,14 +72,52 @@ def _request(
...
+                guidance = "Rate limit exceeded. Wait before retrying."
+            elif error.code >= 500:
+                guidance = "Gangway service failure. Check OpenShift CI service availability."
+            else:
+                guidance = "Check the request and OpenShift CI service availability."
+            raise GangwayAdapterError(
+                f"Gangway returned HTTP {error.code}. {guidance}",
+                outcome_unknown=method == "POST" and (error.code == 408 or error.code >= 500),
+            ) from error
+        except (urllib.error.URLError, OSError) as error:
```
gustavolira, line 101: http.client.HTTPException is not an OSError, so IncompleteRead or BadStatusLine escapes here as a raw traceback. I reproduced it: if a POST succeeds but the connection drops mid body, the job exists and the user never gets the outcome_unknown warning this PR added for exactly that case. Adding http.client.HTTPException to this except (with outcome_unknown for POST) closes it.

skills/ci/rhdh-prow-trigger/scripts/gangway_adapter.py
```diff
@@ -59,14 +72,52 @@ def _request(
...
+                guidance = (
+                    "Authentication was rejected. Run /setup-rhdh-skills openshift-ci, then retry."
+                )
+            elif error.code == 403:
+                guidance = "Permission denied. Ask an OpenShift CI administrator to check access."
+            elif error.code == 400:
+                guidance = "Invalid request. Check the job name, execution ID, and overrides."
+            elif error.code == 404:
+                guidance = (
+                    "Execution not found. Check the execution ID."
```
gustavolira, line 87: Heads up from testing against the live Gangway: it returns 500 for unknown execution IDs, not 404 (tried a garbage ID and a well formed UUID, both 500). So this branch never fires in practice, and a typo in --status gets diagnosed as a service outage. Also 408 is marked outcome_unknown below but falls into the generic "check the request" guidance. Worth a code comment at least, so nobody trusts the 404 path later.

skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
```diff
@@ -96,10 +113,24 @@ def fetch_configured_jobs(repo: str) -> list[str]:
...
     # Extract job names from embedded JSON: "name":"periodic-ci-...-nightly"
-    matches = re.findall(r'"name":"(periodic-ci-[^"]*-nightly)"', html)
+    matches = re.findall(r'"name"\s*:\s*"(periodic-ci-[^"]*-nightly)"', html)
     return sorted(set(matches))
 
 
+def ensure_configured_job(job: str) -> None:
+    """Require membership in the owning repository's live job list before a POST."""
+    repo = REPOS[1] if job.startswith(OVERLAY_JOB_PREFIX) else REPOS[0]
+    jobs = fetch_configured_jobs(repo)
```
gustavolira, line 123: fetch_configured_jobs only catches URLError/OSError, so a proxy or captive portal returning a non UTF-8 or truncated body makes this gate blow up with a traceback instead of the clean "no job was submitted" message. Still fails closed, but the promise of a stated reason is lost. The tests stub fetch_configured_jobs wholesale, so nothing catches this.

skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
```diff
@@ -64,6 +65,22 @@
 ]
 
 OVERLAY_JOB_PREFIX = "periodic-ci-redhat-developer-rhdh-plugin-export-overlays-"
+NIGHTLY_JOB_PATTERN = re.compile(
+    r"periodic-ci-redhat-developer-rhdh-(?:plugin-export-overlays-)?"
+    r"[a-zA-Z0-9][a-zA-Z0-9_.-]*-nightly"
```
gustavolira, line 70: This also matches sibling repos: periodic-ci-redhat-developer-rhdh-operator-main-e2e-nightly passes fullmatch, and the dry run stamps job_name_and_flags: passed for a job the live gate will always reject. Anchoring the branch segment (main or release-N.N) would close it.

skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
```diff
@@ -96,10 +113,24 @@ def fetch_configured_jobs(repo: str) -> list[str]:
...
+    matches = re.findall(r'"name"\s*:\s*"(periodic-ci-[^"]*-nightly)"', html)
     return sorted(set(matches))
 
 
+def ensure_configured_job(job: str) -> None:
+    """Require membership in the owning repository's live job list before a POST."""
+    repo = REPOS[1] if job.startswith(OVERLAY_JOB_PREFIX) else REPOS[0]
+    jobs = fetch_configured_jobs(repo)
+    if not jobs:
+        log_error(f"Cannot verify configured nightly jobs for {repo}; no job was submitted.")
```
gustavolira, line 125: Question on the fail closed choice: this is a single attempt scrape of the Deck HTML, no retry, no bypass. If the page hiccups or its format changes, every trigger is blocked, and release week (new release branch jobs, RC verification) is when that hurts most. I checked the scrape works today (53 rhdh + 2 overlay jobs found), so this is about resilience, not correctness. Would you take a retry plus an explicit skip flag for emergencies? Also this runs before ensure_capability, so someone with an expired session waits for the fetch before hearing they need to log in.

skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
```diff
@@ -577,16 +639,27 @@ def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
...
-        log_error("Either --list, --list-tags, or --job is required.")
+    if args.dry_run and not args.job:
+        log_error("--dry-run requires --job.")
         sys.exit(1)
 
+    if args.status is not None:
+        if not re.fullmatch(r"[a-zA-Z0-9][a-zA-Z0-9_-]*", args.status):
+            log_error("--status requires an execution ID, not a URL or path.")
+            sys.exit(1)
+        if any(getattr(args, name) for name in TRIGGER_OVERRIDES) or args.tag_filter:
```
gustavolira, line 650: nit: --status rejects every trigger flag loudly but --json slips through and does nothing (show_job_status ignores json_output). Either reject it too or honor it, the silent accept is the worst of the two options.

skills/ci/rhdh-prow-trigger/workflows/trigger-nightly.md
```diff
@@ -11,9 +11,12 @@ Run every command below from this skill's directory.
...
 ## Flow
 
-1. Fetch available jobs and let the user pick one
-2. Ask about image override and additional options (fork, alerts)
-3. Show the command, confirm, execute, report results
+1. Resolve the requested job; fetch available jobs when selection is needed
+2. Fill in any requested overrides that are missing values
+3. Preview, apply the write gate, execute, and report results
+
+For a supplied full job name and flags, go directly to the preview in Step 3.
```
gustavolira, line 18: This shortcut skips Step 1, which is the only place in this workflow that carries the GKE/OSD-GCP shared cluster warning. Someone pasting a full gke job name is exactly the case that needs it. Maybe repeat the warning in Step 3?

skills/ci/rhdh-prow-trigger/workflows/trigger-nightly.md
```diff
@@ -145,10 +148,38 @@ uv run scripts/trigger_nightly_job.py \
...
+
+### Status and recovery
+
+To refresh an existing execution, use the GET-only status command:
+
+```bash
+uv run scripts/trigger_nightly_job.py --status <EXECUTION_ID>
+```
+
+The CLI also prints this command after submission. Repeating a trigger creates
```
gustavolira, line 165: nit: two small mismatches with the script here. The refresh command is only printed when Gangway returns an id (the no id branch says --status is unavailable instead), and the incompatibility list misses --tag-filter, which validate_args also rejects.

tests/unit/test_trigger_nightly_job.py
```diff
@@ -55,19 +85,258 @@ def fake_run(argv, **_kwargs):
...
+    with pytest.raises(SystemExit) as error:
+        NIGHTLY.main(["--job", job, *(["--dry-run"] if dry_run else [])])
+    assert error.value.code != 0
+
+
+@pytest.mark.parametrize(
+    "extra",
+    [
+        ["--job", JOB],
+        ["--dry-run"],
```
gustavolira, line 159: This case exits through the "--dry-run requires --job" guard, not the status conflict branch the test is named after (I checked; same story for the empty job plus dry_run case in test_unsupported_job_names_fail_offline). Asserting the stderr message would pin the right branch. Related gap: the HTTP error tests only drive trigger(), so the GET arm (404 message, outcome_unknown False) is untested, and nothing pins the poll loop attempt count in either direction.

gustavolira:
Took a deeper look since the question was whether these concerns are real or noise. They're real. I checked the three claims against main and all of them hold: the "Re-run this command to refresh" hint really would submit a duplicate ProwJob, the old validation accepted any periodic-ci job name, and every API error got blamed on the job name.

Beyond reading the diff I also validated it against the real systems: ran the full suite locally (673 pass, matches the PR), pulled the live configured-jobs pages (53 rhdh + 2 overlay jobs, all accepted by the new pattern, so no legit job gets rejected), replayed the release-1.10 dry run (payload identical to main), and made some read only calls to the real Gangway with my CI session. That last part is where the interesting findings came from: Gangway answers 500 for unknown execution IDs, which changes how the new error taxonomy and the polling behave in practice. Details in the inline comments.

Overall the direction is good and the tests are honestly above average, the autouse fixture that fails on any unmocked network call is a nice touch. My asks before merge are the polling early return, the uncaught http.client exceptions in the adapter, and a decision on the fail closed scrape gate (retry or escape hatch).

Smaller things I didn't bother making inline comments for: the REPOS[0]/REPOS[1] indexing in ensure_configured_job breaks silently if the list is reordered, TRIGGER_OVERRIDES means a new flag now needs three coordinated edits, and the bare invocation exit code changed from 1 to 2 with the required group. None of them blocking.

skills/ci/rhdh-prow-trigger/scripts/trigger_nightly_job.py
nickboldt: retry how many times / how long? hammering the server with "u up?" messages is how we get 500s in the first place: it's busy with the DDOS attacks from other clients asking "u up?" ;)

gustavolira: @nickboldt I’m not asking for more requests, just smarter retries.

Keep the existing 5-GET cap, but use backoff (1s, 2s, 4s, 8s) and fail immediately on 401/403. Retry only 5xx/network errors.

That matters because Gangway can return 500 for a newly submitted ID, so today a transient first response can kill polling even when the job is fine. Since --status <ID> already exists for manual refresh, the loop doesn’t need to wait longer

nickboldt:
since we're ai-screening this, perhaps we can use ai to refactor this using your suggestions? WDYT @gustavolira ? You're the SME here, so maybe we close this and replace with something more aligned to your suggestions? 

(this was ALSO  an experiment to see how crappy the output is from `gpt-6-astra xhigh` vs. Cursor w/ grok)

nickboldt:
based on the AI-generated review of this AI-slopped code, I'll just close this.

PR #117, opened 2026-09-15
Add rhdh-release-sos skill for timeline-aware Scrum of Scrums release check-ins

psrna:
## Summary

Introduces **`/rhdh-release-sos`**, a read-only Agent Skill that builds the fixed **Scrum of Scrums (SoS) release check-in** report for RHDH team leads. It answers “where are we on the release calendar, which checklist items are due today, and what do the Jira counts look like?” — without ad-hoc JQL or manual table assembly.

The skill is intentionally separate from **`/rhdh-release-status`** (flexible status queries) and **`/rhdh-release-schedule`** (milestone dates alone). SoS runs a curated runbook in `references/sos-checks.md`: checks appear only in the active release phase (Feature Freeze, Code Freeze, Go/No Go, GA Announce) and only once their offset date is on or before the as-of date. Jira execution reuses the release-status Rich Filter stack via `scripts/_release_bridge.py`.

Deliverables on `run`:

- **`report_markdown`** — fixed template (metadata, milestones, checks with team sub-rows)
- **Self-contained HTML** — same data, highlighted checks, collapsible team breakdown, action-item hints on hover
- **`report_html_path`** — timestamped file under `reports/` (or a path from `--html-output`)

Runbook highlights:

- Query kinds: `static`, `due static`, `expect assignee`, Test Plan (`testplan`), epic Dev Complete ratio, templates, and queues
- **Action item** column in `sos-checks.md` (optional per row); markdown uses a `↳ ℹ …` sub-row when set; HTML uses an info icon with a CSS tooltip (blank cells omit both)
- New/updated checks including FF scope not in sprint, unassigned Stories/Tasks at FF −21d, Test Day owner, Test Plan assignment and sign-off, due-by-milestone work remaining, and phase snapshots

Also registers the skill in the pack catalog and adds unit tests for parsers, check selection, formatters, and report helpers.

## What this skill is for

Use **`/rhdh-release-sos`** when preparing or running the **recurring SoS release check-in**: paste the CLI markdown into the meeting and share the HTML for leads who want Jira links and team breakdowns. It does **not** post to Jira or Slack; it only reports counts and links from the operational Rich Filter export.

Example triggers: “SoS report for 2.1.0”, “scrum of scrums release status”, “release check-in for team leads”, “what should we review before feature freeze”.


skills/release/rhdh-release-sos/scripts/sos_checks.py
```diff
@@ -0,0 +1,218 @@
...
+
+    if as_of < ordered[0][1]:
+        return ordered[0][0]
+
+    for index in range(len(ordered) - 1):
+        start, end = ordered[index][1], ordered[index + 1][1]
+        if start <= as_of < end:
+            return ordered[index + 1][0]
+
+    return ordered[-1][0]
```
schultzp2020: `start <= as_of < end` advances the phase *on* the milestone day, so the `-0d` snapshot rows can never run: on Feature Freeze day (2026-09-22) the phase resolves to Code Freeze and the due list is empty — reproduced with the real runbook; Code Freeze day same. `sos-checks.md` promises the opposite ("Row can still appear if its `When` date is due (e.g. FF - 0d snapshot)"), and `references/timeline.md`'s interval table encodes the current (broken) mapping. Milestone day should close the section it belongs to — this also needs the interval table in `timeline.md` updated to match, and ideally a milestone-day test:

```suggestion
    if as_of <= ordered[0][1]:
        return ordered[0][0]

    for index in range(len(ordered) - 1):
        start, end = ordered[index][1], ordered[index + 1][1]
        if start < as_of <= end:
            return ordered[index + 1][0]

    return ordered[-1][0]
```

skills/release/rhdh-release-sos/scripts/sos_checks.py
```diff
@@ -0,0 +1,218 @@
...
+        )
+
+    if not rows:
+        raise ValueError("No SoS checks found in sos-checks.md")
+    return rows
+
+
+def load_checks(base: Path | None = None) -> list[SectionHeader | CheckRow]:
+    path = checks_path(base)
+    return parse_checks(path.read_text())
```
schultzp2020: `read_text()` uses the locale default encoding; `sos-checks.md` contains non-ASCII dashes and ellipses, so on Windows (cp1252) `load_checks()` dies with `UnicodeDecodeError: byte 0x9d` — three tests in `test_sos_checks.py` fail on Windows while CI (UTF-8 locale) stays green. Verified: with the encoding pin, 3 failed → 10 passed.

```suggestion
    return parse_checks(path.read_text(encoding="utf-8"))
```

skills/release/rhdh-release-sos/scripts/sos_report.py
```diff
@@ -0,0 +1,712 @@
...
+            issues = []
+        else:
+            signoff_jql = testplan_mod.signoff_tasks_jql(epic_key)
+            signoff_tasks = release_mod._acli_json_enriched(
+                signoff_jql,
+                select="key,summary,status,assignee",
+                limit=testplan_mod._CHILD_SEARCH_LIMIT,
+            )
+            evaluation = testplan_mod.evaluate_signoff_open(signoff_tasks)
+            jira_url = epic["epic_url"] or jql_mod.jira_url(testplan_mod.signoff_open_jql(epic_key))
```
schultzp2020: With open sign-off tasks this links the Jira column to the epic, while the sibling children check links the actionable *unassigned filter* (`unassigned_children_jql`, line 256). Leads following the link land on the epic instead of the open sign-off tasks. Point at the open filter while work remains:

```suggestion
            jira_url = (
                jql_mod.jira_url(testplan_mod.signoff_open_jql(epic_key))
                if evaluation.get("open_count")
                else epic["epic_url"]
            )
```

skills/release/rhdh-release-sos/scripts/sos_report_format.py
```diff
@@ -0,0 +1,817 @@
...
+        dt = dt.replace(tzinfo=timezone.utc)
+    return dt.astimezone(timezone.utc).strftime("%Y-%m-%d %H:%M:%S UTC")
+
+
+def report_as_of_display(report: dict) -> str:
+    """Human-readable As of value for rendered reports."""
+    generated_raw = report.get("generated_at")
+    if generated_raw:
+        return format_generated_at(generated_raw)
+    return f"{report['as_of']} 00:00:00 UTC"
```
schultzp2020: `report_as_of_display` prefers `generated_at`, so `run 2.1.0 --date 2026-09-08` renders "As of: <generation timestamp>" in both markdown and HTML — the requested as-of date never shows. Is generation-time display intended (it matches same-day runs), or should the rendered metadata show `as_of`, with generation time left to the JSON's `generated_at` field?

skills/release/rhdh-release-sos/scripts/_release_bridge.py
```diff
@@ -0,0 +1,31 @@
+"""Reuse query and date helpers from the sibling rhdh-release-status skill."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+
+_STATUS_SCRIPTS = Path(__file__).resolve().parents[2] / "rhdh-release-status" / "scripts"
```
schultzp2020, line 8: This reaches into release-status internals via `sys.path` (`_acli_count`, `_acli_json_enriched`, `_fetch_teams`, `_extract_milestone_dates`), so SoS breaks when installed standalone and couples to private helpers. ADR-0006 allows code duplication precisely so a skill installs alone, but `catalog.json` declares `requiresSkills: [rhdh-release-status, rhdh-release-schedule]` — is the contract "install both, fail loudly" (current) or "self-contained"? If the former, worth one sentence in SKILL.md's `compatibility`; if the latter, copy the helpers per ADR-0006.

schultzp2020 (changes requested):

skills/release/rhdh-release-sos/scripts/_release_bridge.py
psrna: its the "install both, fail loudly", I will add a note to the skill.md's compatibility as you are suggesting. thx

skills/release/rhdh-release-sos/scripts/sos_report_format.py
psrna: Good catch, not intentional for --date runs. Will fix.

skills/release/rhdh-release-sos/scripts/_release_bridge.py
psrna: text updated.

schultzp2020 (approved):
LGTM!

